PR #3, opened 2018-03-27
use eslint

sastels:
after the CI PR merges will bring master into this one.


maxneuvians (changes requested):
CI fails - does it pass the test locally?

sastels:
It fails because this branch doesn't have the circleci config stuff, but circleci is trying to run anyway :) should be fixed by merging the new master in.

maxneuvians:
then gtg!

Issue #10, opened 2018-03-28
Prettier


maxneuvians:
Use https://prettier.io/ for code formattig

PR #11, opened 2018-03-28
Prettier config and run on pre-commit

maxneuvians:
Adds a prettier dependency and as well as a runner to run prettier on touched file before a commit. Maybe conflict with ESLinting rules, ex: Prettier uses " convention for strings vs. ' which certain styles prefer. #10


sastels (approved):
👍

PR #24, opened 2018-04-03
deploy only when build passes and on master branch

szinck1:
This PR adds a workflow to the CI so that `deploy` only runs when `build` passes. `deploy` will also only run on `master`.


maxneuvians (approved):
👍

sastels (approved):
:shipit:

Issue #20, opened 2018-03-29
Ensure IE compatability


maxneuvians:
See https://github.com/cds-snc/nrcan-energuide-api-poc/issues/51

sastels:
For IE9 we'll probably need
https://reactjs.org/docs/javascript-environment-requirements.html

sastels:
Supporting IE11 is all that is required, so we're already fine.

We should remove the `core.js` and `ref` lines / packages.

PR #25, opened 2018-04-03
Use core-js

sastels:
re #20, import `core-js` to make things work on Internet Explorer 9 and 10?


maxneuvians (approved):
Not sure if this solves our IE / Next.js issues but will need this irregardless for IE < 11. 👍

PR #33, opened 2018-04-04
Added Sentry HOC

maxneuvians:


lib/withSentry.js
```diff
@@ -0,0 +1,33 @@
+import React from "react";
+import Raven from "raven-js";
+
+const SENTRY_DSN = "https://b23c49f878524c8eaf98fb47340d2448@sentry.io/1082251";
```
sastels: This isnt some sort of key, right?

sastels (approved):
Minor question, :shipit:

lib/withSentry.js
maxneuvians: It is the public portion of the key, ideally this would be injected through ENV variables but it would still be in the bundle served to the client. All it does is prevent it from being embedded in other domains and it is up to us in sentry to configure the correct domains to track from. 
screen shot 2018-04-05 at 7 41 46 am

maxneuvians:
Configured the Sentry Public key as part of the CircleCI ENV variables

PR #40, opened 2018-04-05
remove core-js and raf

### LINKED ISSUE #20: Ensure IE compatability

sastels:
fixes #20 

We don't need this code since we don't need to support IE < 11 (and we couldn't get it to work anyway)


maxneuvians (approved):
🎃

PR #44, opened 2018-04-05
Inject SENTRY_DSN and GA_UA into Docker build

maxneuvians:
@szinck1 Could you take a look just to make sure if this makes sense? Both SENTRY_DSN and GA_UA are configured in Circle_CI. The idea is that they get passed to the node build process and then get written into the compiled code instead of their placeholders.


sastels (approved):
👍

PR #49, opened 2018-04-10
Cypress E2E testing

### LINKED ISSUE #31: Investigate E2E testing with cypress.io
https://www.cypress.io/

sastels:
resolves #31 

local testing via `yarn cypress:open` and `yarn cypress:run`. Also running in CI.

### DIFF
--- a/cypress/fixtures/example.json
+++ b/cypress/fixtures/example.json
@@ -1,0 +1,5 @@
+{
+  "name": "Using fixtures to represent data",
+  "email": "hello@cypress.io",
+  "body": "Fixtures are a great way to mock data for responses to routes"
+}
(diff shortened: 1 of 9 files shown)


cypress/fixtures/example.json
```diff
@@ -0,0 +1,5 @@
+{
```
maxneuvians, line 1: Do we need this file?

maxneuvians (approved):
LGTM

cypress/fixtures/example.json
sastels: It's basically the file where you put your fixtures, this example content was created on first run of cypress. Down the line we'll most likely replace it with real fixtures.

PR #57, opened 2018-04-11
Layout

maxneuvians:
Added a layout component that wraps the head and errorBoundary


sastels (approved):
As far as I can see everything is working, esp wrt i18n: common is being picked up, and `t` is working in `Head` 👍

Issue #69, opened 2018-04-13
Page A3


sastels:
https://app.zeplin.io/project/5acbd659dc553ec84805afd4/screen/5ace615f84aaee9f55e22a3e

![image](https://user-images.githubusercontent.com/8228248/38746483-96cb9c7c-3f15-11e8-9ad6-e1bc81b81957.png)

sastels:
Integrate with real data from data store: Wait until #75 is completed

PR #73, opened 2018-04-16
Benefit title card

sastels:
in support of #69 

Add a title card component (along with a list of cards component) for benefits.
image


maxneuvians (approved):
Looks great!

PR #74, opened 2018-04-16
Css fixes

maxneuvians:
Removed the old SCSS stylesheets. They did not make sense once the styling comes largely from material UI and are in-lining styles. To keep with the current trends we can merge css into the components using emotion and react-emotion:

https://emotion.sh/

I removed the old SCSS stuff which worked better with Bootstrap. Look at component/footer.js for an example.


sastels (approved):
:octocat:

PR #79, opened 2018-04-17
Filters selected card

sastels:
In support of #69 
A card with a given list of filters that the user has selected.

image


maxneuvians (approved):
GTG 🦑

PR #80, opened 2018-04-17
Fix floating footer

sastels:
Stop footer from floating when the screen scrolls, ie

image

Also delete the sum() example code


maxneuvians (approved):
LG

Issue #51, opened 2018-04-11
Look into SSR for Material


maxneuvians:
https://material-ui-next.com/guides/server-rendering/

PR #81, opened 2018-04-17
CSS SSR

maxneuvians:
Added ability for the app to render the CSS on the server and put in with initial package. Both Emotion CSS and Material UI CSS are injected in `_document.js`. The `Layout` component will remove the CSS upon render.


sastels (approved):
🐙

sastels:
Does this resolve #51  ?

PR #96, opened 2018-04-18
Added fallback pages for IE < 11

maxneuvians:
Added a fall back page when the server encounters a IE browser less than IE < 11. Instead a static html page is sent back to the browser. Added another yarn option to specifically test cypress with a IE 10 user agent string.


sastels (approved):
:octocat:

PR #97, opened 2018-04-18
Use redux

### LINKED ISSUE #75: Integrate Data Store Into App
Are there any related nextjs best practices that should be observed in terms of what we chose, or how it is configured?

sastels:
Resolves #75 
Initial code to make a redux datastore and use it for the benefit data.


maxneuvians (approved):
Looks good!

PR #100, opened 2018-04-18
Footer fix #95

### LINKED ISSUE #95: Fix floating footer
The footer no longer overlaps the content, but if the page contents is too short it's in the middle of the browser :(

image

maxneuvians:
Made some small CSS adjustment to fix footer


sastels (approved):
🦀

PR #102, opened 2018-04-18
Init layout for A3

sastels:
re #69 

Initial code for a layout (hardcoded values, no actual filtering)

image


maxneuvians (approved):
Great!

PR #105, opened 2018-04-19
Added alpha banner

maxneuvians:
Added an alpha banner to make sure people know we are in Alpha. To be replaced with Beta banner. Probably never to be replaced by a Delta Banner.


sastels:
What about a Gamma Banner?

sastels (approved):
A

maxneuvians:
The Gamma quadrant was always a hive of scum and villainy *cringe* @danprime

PR #106, opened 2018-04-20
Use Airtable

### LINKED ISSUE #88: Hydrate initial with all data required for page A3

sastels:
resolves #88 and #69 
* Hydrate Redux from Airtable.
* change the pages to use the Airtable schema
* filter benefits on A3 according to choices made on A1 and A2.

image


package.json
```diff
@@ -33,6 +33,7 @@
     "material-ui": "next",
     "next": "^5.1.0",
     "next-redux-wrapper": "^1.3.5",
+    "npm": "^5.8.0",
```
maxneuvians: Why do we need NPM?

maxneuvians (commented):
Just need to double check if we need NPM otherwise GTG!

maxneuvians (approved):
🦑

PR #108, opened 2018-04-20
just use prefix of gitsha in footer

sastels:
otherwise the footer is too wide for mobile


sastels:
Need to make the alpha message shorter too

maxneuvians (approved):
:squid: 🦈

maxneuvians:
Can also mess around with CSS - https://www.w3schools.com/cssref/css3_pr_word-wrap.asp to break the message onto two lines, but shortening should also work.

PR #111, opened 2018-04-23
fix some translation related issues

sastels:
* fixed a bug in how `SelectButton` was being used by the `Next` and `See Results` buttons
* reworded the "no benefits" message on A3


maxneuvians (approved):
LGTM 🦑

PR #113, opened 2018-04-23
Redo SelectButton tests

sastels:
following approach in https://medium.freecodecamp.org/the-right-way-to-test-react-components-548a4736ab22


sastels:
Just noticed the "Top level is a Card" test isn't correct, will fix...

maxneuvians (approved):
🦑

PR #119, opened 2018-04-24
Added lang attribute to html

maxneuvians:
HTML requires a lang attribute. The app will now add either `en` or `fr` depending on the initial browser language.


sastels (approved):
🎸

PR #122, opened 2018-04-25
Single page app

### LINKED ISSUE #94: Make A option a single page app
The "Next" button on A1 is currently just an `href` so it hits the server to get page A2.
image

sastels:
resolves #94 

* Put `A1`, `A2` and `A3` into a page `A`
* add navigation buttons from `A3` back to `A1` and `A2`
* maintain selected options while moving around between `A1`, `A2` and `A2`
* cleanup some unused code and add `<h1>`'s


maxneuvians (approved):
🦑

PR #123, opened 2018-04-25
start with empty datastore

sastels:
It's confusing to have a few options when page loads, then more appear after AirTable loads.


maxneuvians (approved):
LG

PR #125, opened 2018-04-26
Fix BenefitCard

### LINKED ISSUE #124: Fix BenefitCard
switching to airtable's schema broke the (not currently used) BenefitCard (a card that showed the one line description of the benefit)

sastels:
fixes #124 

The `BenefitCard` that contains the one line description broke when we switched to using the Airtable schema in our datastore. This fixes that and uses those cards on the `all-benefits` page.

Also minor fixes to the header and footer (eliminates unnecessary scroll bars)


maxneuvians (approved):
LGTM 🦑

PR #130, opened 2018-04-26
Use jest coverage

### LINKED ISSUE #114: investigate code coverage tools
ex: https://github.com/istanbuljs/babel-plugin-istanbul

sastels:
resolves #114 

set thresholds at 50% so that we pass. Will increase next week when we improve the tests.


maxneuvians (approved):
🦑

PR #131, opened 2018-04-27
open Show All Benefits in new tab

sastels:
We probably want to do this, otherwise (currently) you lose your state when you go back to page A.


maxneuvians (approved):
LGTM!

PR #132, opened 2018-04-27
bookmarking of state in url

### LINKED ISSUE #85: Bookmarking : user can return to a certain filter state of the app
With a URL is preferred

sastels:
resolves #85 

Put info in the url when changing sections, and parse it out when the page loads.
```
http://localhost:3000/A?section=A3&selectedBenefitTypes=recQO4AHswOl75poF&selectedPatronTypes=rec726lY5vUBEh2Sv,recDAuNt8DXhD88Mr
```

### DIFF
--- a/__tests__/pages/A_test.js
+++ b/__tests__/pages/A_test.js
@@ -13,6 +13,7 @@ describe("Page A", () => {
   it("Layout", () => {
     const app = mount(
       <App
+        url={{ query: {} }}
         i18n={{ language: "en-US" }}
         t={key => key}
         storeHydrated={true}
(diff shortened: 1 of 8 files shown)


maxneuvians (commented):
We should test this change with cypress?

sastels:
good Idea. 👍

sastels:
ugh, before we should be using Cypress for E2E testing we need to get Cypress to mock the Airtable calls.

maxneuvians:
Nice LGTM 🦑

sastels:
right now using Cypress to test navigation around the app. Will add Airtable mocking in a different PR.

maxneuvians (approved):

Issue #135, opened 2018-05-01
Rewrite and Consolidate Around an Approach to Unit Tests for all non-A components


leonL:
Rewrite all tests in app as a pair, and by doing so define a standard methodology for how we will approach testing in the app going forward.

*Consider discussing react testing strategy with other CDS devs

- [x] alpha_banner.js			
- [x] footer.js			
- [x] menu_bar.js			
- [x] select_button.js
- [x] benefit_cards.js		
- [x] head.js				
~~nav.js~~			
- [x] selected_options_card.js		
- [x] errorBoundary.js		
- [x] layout.js			
- [x] noscript.js

sastels:
Woot!

PR #145, opened 2018-05-02
Reorg benefit cards and test

sastels:
re #135
move List code into A3 and all-benefits and test BenefitCard and BenefitTitleCard.


maxneuvians (approved):
Looks good! 🦑

PR #152, opened 2018-05-03
Ignore _document.js in jest

sastels:
will test _document.js with Cypress


maxneuvians (approved):
Ignored test is best test 🦑

PR #154, opened 2018-05-04
Cypress typical story

sastels:
* if url param `use_testdata` is set, load data from fixtures rather than AirTable
(example: http://localhost:3000/A?use_testdata=true)
* use this in Cypress testing to test a typical user interaction story
* rename `fetchFromAirtable` to `hydrateFromAirtable`

### DIFF
--- a/__tests__/pages/A_test.js
+++ b/__tests__/pages/A_test.js
@@ -45,19 +45,19 @@ describe("A", () => {
     _mountedA = undefined;
   });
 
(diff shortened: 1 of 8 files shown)


maxneuvians (approved):
I like this approach. It reminds of the approach here: http://blog.plataformatec.com.br/2015/10/mocks-and-explicit-contracts/ because we are specifiying a contract between the app and the data, not the app and the source of the data. LGTM 🦑

PR #163, opened 2018-05-16
Cache Airtable data in server

sastels:
Cache the AirTable data on the server, refreshing it every hour. Serve this data to the client on page request. (So now the clients never hit AirTable)

in page A, the page gets rerendered when we push changes to the url (just rerendered, not reloaded from the server). To keep the data around between rerenderings we store it in Redux. This causes the same flicker on initial load as we say when the browsers were hitting airtable themselves, unfortunately.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -42,7 +42,7 @@ jobs:
   integration:
     docker:
       # the Docker image with Cypress dependencies
-      - image: cypress/base:6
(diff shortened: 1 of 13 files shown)


sastels (commented):
🦑

maxneuvians (approved):
🦑 Looks good since we mostly did this together

PR #166, opened 2018-05-18
Use new airtable data

sastels:
Use new AirTable eligibility data in a new version of the B page. 
* load data benefits and eligibility data from new tables
* add filters for new eligibility criteria


maxneuvians (approved):
Buffs out 🦑

Issue #169, opened 2018-05-22
Elaborate the benefit cards to show parent/child relationships and specific features of the benefit


leonL:

leonL:
Follow up with Raluca to find out if placeholder data for features related to benefits will be available by the end of the sprint.

sastels:
An initial version of this is done.

PR #179, opened 2018-05-23
Show child benefits

sastels:
re #169 

Add a child card to benefits that have child benefits. We still need to improve both the parent and child cards.


smcmurtry:
looks good.

smcmurtry (approved):

PR #185, opened 2018-05-23
spiff up parent cards a bit

sastels:
- move details button to upper right corner.
- add text if there are child benefits


maxneuvians (approved):
🦑 Looks good!

sastels:
Resolves #181

PR #187, opened 2018-05-23
Add translations for checkbox labels

sastels:
They should really be in AirTable, but for now I'll just put them in the i18n tables


maxneuvians (approved):
Looks good 🦑

PR #190, opened 2018-05-24
Don't show main cards for benefits that need a gateway

sastels:
If a benefit requires a gateway benefit, only show it as a child benefit, not as a main card.

Also, added a column to AirTable benefits table to support this.


smcmurtry (commented):
Looks good

smcmurtry (approved):

maxneuvians (approved):
🦑

Issue #180, opened 2018-05-23
Fix BB tests and refactor the whole thing


maxneuvians:
BB Component is a bit weak in the knees on testing and could use another pass

maxneuvians:
:fire: https://github.com/cds-snc/vac-poc/pull/246/files#diff-108dd3482227c24e297c872ac544fee1R201

sastels:
woot!

PR #192, opened 2018-05-24
Add disclosure for list of child benefits

sastels:
In the parent card, the list of child benefits are now initially hidden by a disclosure button

### DIFF
--- a/__tests__/components/benefit_card_test.js
+++ b/__tests__/components/benefit_card_test.js
@@ -83,7 +83,15 @@ describe("BenefitCard", () => {
     props = {
       t: () => "en",
       benefit: benefitsFixture[0],
-      allBenefits: benefitsFixture
+      allBenefits: benefitsFixture,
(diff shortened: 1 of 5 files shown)


__tests__/components/benefit_card_test.js
```diff
@@ -88,7 +88,15 @@ describe("BenefitCard", () => {
     props = {
       t: () => "en",
       benefit: benefitsFixture[0],
-      allBenefits: benefitsFixture
+      allBenefits: benefitsFixture,
```
maxneuvians, line 91: Don't think we actually need these? Don't see them referenced in the test

maxneuvians (approved):
Looks good, will address again in #180 🦑

__tests__/components/benefit_card_test.js
sastels: needed in the BenefitCard code

PR #197, opened 2018-05-24
Fixes to bcard

maxneuvians:
Fixes 

Add box to child card and indent more
Remove “”these benefits” from above child cards
Make “View Detail” buttons more obvious
Look into why “With this benefit, you may…,” overflows cards sometimes
Change “Benefit Description” on French page
 Make one columns of benefits


smcmurtry (approved):
I like it

PR #200, opened 2018-05-25
change benefit filtering logic

sastels:
re conversations in Slack, change the way we filter the benefit cards:
- show main cards for a benefit even if it appears as a child card
- show all child cards of a gateway benefit, even if some of those child cards do not match the filters.


maxneuvians (approved):
🦑

PR #201, opened 2018-05-25
stop child card disclosure from floating to right of card

sastels:
Now that cards are wider it looks funny to have the disclosure all the way to the right.

We might want to allow the user to click on the line of text as well and use an accordion disclosure.

### DIFF
--- a/components/benefit_cards.js
+++ b/components/benefit_cards.js
@@ -22,10 +22,6 @@ const buttonStyles = {
   marginRight: "25px"
 };
 
-const chevronStyle = {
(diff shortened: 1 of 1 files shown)


maxneuvians (approved):
I think the intial idea was to have it act more like this

https://material-ui.com/demos/cards/#complex-interaction

But I agree that a whole clickable line with a different bg colour might be the better solution

sastels:
@maxneuvians indeed, right now it behaves as in the material-ui  demo (that's where I stole the code from :O )

PR #203, opened 2018-05-25
target Dan in all-benefits link

sastels:
So all benefits page will only open open tab if reclicked


locales/en/common.missing.json
```diff
@@ -1,3 +0,0 @@
-{
```
sastels, line 1: Lol

smcmurtry (approved):
I like it.

PR #206, opened 2018-05-25
VAC WG branch - not a real PR

sastels:
Don't merge this, it's just so if we change the WG branch we can check it out on Heroku before we redeploy to vac-poc app.


sastels:
Don't need this as a PR anymore

PR #212, opened 2018-05-28
Put filters in url

### LINKED ISSUE #210: Add filters to URL
Similar to what we had before for the "A" option, 
- [x] add selected filters to the url
- [x] add a url parser that runs when page is loaded
- [x] add new url to browser's stack so that back / forward buttons work

sastels:
resolves #210 

* add selected filters to the url
* add a url parser that runs when page is loaded
* make browser back / forward buttons work


maxneuvians (approved):
:squid: look good - there has to be a native `stringToMap` somewhere in JS land, but your implementation looks great

Issue #209, opened 2018-05-28
Left border of card should change depending on whether card is open / close


sastels:
re https://projects.invisionapp.com/freehand/document/aqX9F1ne1

Add a line down the left hand side of the card showing whether the card is open or not. Alternatively, do something that does a s good a job of subtly showing which cards are open or closed.

maxneuvians:
Closed by #221

PR #221, opened 2018-05-29
Added Left border on benefits cards

maxneuvians:
Feature for #209


sastels:
Add to child cards too?

smcmurtry (approved):

Issue #208, opened 2018-05-28
Refine child benefit card


sastels:
Update UI of child card to bring it in line with ( https://projects.invisionapp.com/freehand/document/aqX9F1ne1). look at things like: change the "View Details" button to a "Find out more" link AND  add "Available to" data (to AirTable and app) AND use your judgement or we will replace you with a machine learning algorithim

smcmurtry:
"Available to" data still needs to be added to AirTable and app. The other stuff is done.

PR #225, opened 2018-05-29
Refine child benefit

smcmurtry:
Did everything in #208 except the "available to" bullet points are lorem ipsum rather than real data


maxneuvians (approved):
Looks good 🦑

PR #227, opened 2018-05-29
Use a polyfill to fix ie11

### LINKED ISSUE #218: App is broken on IE11

sastels:
fixes #218 

Use a babel polyfill ( https://babeljs.io/docs/usage/polyfill/ )

This seems to do the trick! We'll have to test IE compatibility on a regular basis...


pages/A.js
```diff
@@ -31,10 +32,19 @@ export class A extends Component {
...
+    let hashes = search.slice(search.indexOf("?") + 1).split("&");
+    return hashes.reduce((params, hash) => {
+      let [key, val] = hash.split("=");
+      return Object.assign(params, { [key]: decodeURIComponent(val) });
+    }, {});
+  };
+
   componentWillMount() {
     Router.onRouteChangeStart = newUrl => {
-      const myURL = new URL(newUrl, "http://hostname");
```
sastels, line 36: Alas, there's no `URL` class in IE11 for us to use to parse our url string, so I had to write my own url parser (well, steal one from StackOverflow, but that's how kids code these days, isn't it?)

maxneuvians (approved):
🦑

PR #234, opened 2018-05-30
Next6

maxneuvians:
- Upgraded to Babel7
- Upgraded to NextJS6
- Upgraded react-next-wrapper
- Moved Store initialization logic to _app.js
- Simplified Store
- Removed Flow because of incompatibility with Babel7 and lack of use

**Technical Debt: Add back in prop types**


components/head.js
```diff
@@ -1,18 +1,8 @@
-// @flow
```
sastels, line 1: 🔥 rather than comment out?

components/BB.js
```diff
@@ -1,5 +1,3 @@
-// @flow
```
sastels: 🔥

components/head.js
sastels: nothing to see here, carry on.

sastels (approved):
🦑 🤞

PR #236, opened 2018-05-30
Benefit examples

### LINKED ISSUE #216: add 'examples' to card
- check-in with design about the look of the list before accepting the PR
- consider edge case of there being no or a lot of examples (show more button?)

sastels:
resolves #216 

Add benefit examples to card (if they exist in AirTable). Currently AirTable only has examples for Rehabilitation Services.


maxneuvians (approved):
:squid:

PR #238, opened 2018-05-30
Benefit sorting

smcmurtry:


__tests__/components/BB_test.js
```diff
@@ -74,6 +74,13 @@ describe("BB", () => {
     ).toEqual([benefitsFixture[1]]);
   });
 
+  it("has a correct sortBenefits function", () => {
```
sastels: is this actually testing the sort? shouldn't we add a priority to the fixtures?

components/BB.js
```diff
@@ -171,14 +202,17 @@ export class BB extends Component<Props> {
...
     const { t, classes } = this.props; // eslint-disable-line no-unused-vars
-
     const filteredBenefits = this.filteredBenefits(
       this.props.benefits,
       this.props.eligibilityPaths,
       this.props.selectedEligibility,
       this.props.needs,
       this.props.selectedNeeds
     );
+    const sortedFilteredBenefits = this.sortBenefits(
```
sastels: How about we sort first, and use the sortedBenefits for the children as well?

sastels (approved):
minor suggestions, LGTM

PR #245, opened 2018-05-31
Fix to load all ENV variables in Container

maxneuvians:
This should fix the deploy notification script in Azure


sastels (approved):
sure, let's give it a try

Issue #232, opened 2018-05-30
When to show parents and children


sastels:
- If a parent card is shown, always show all children (never filter children) (current behaviour)
- If child eligible but parent is not, show parent card (need to do)
- If a benefit already appears as a child - show as a main card if and only if it’s also independent

PR #246, opened 2018-05-31
Display benefit criteria

sastels:
resolves issue #232 

- If a parent card is shown, always show all children (never filter children)
- If child eligible but parent is not, show parent card
- If a benefit already appears as a child - show as a main card if and only if it’s also independent


maxneuvians (approved):
🦑 Looks good - I put a reference to to https://github.com/cds-snc/vac-poc/pull/246/files#diff-108dd3482227c24e297c872ac544fee1R201 In the Fix / refactor BB

PR #248, opened 2018-06-01
Collapse button

smcmurtry:
1 thing left undone: it should only be visible if there are expanded benefits


maxneuvians (approved):
Nice! I am happy the `setState` worked out on children 🦑

smcmurtry:
closes #230

PR #251, opened 2018-06-01
Plus minus buttons

smcmurtry:
The button still rotates. Not sure how to stop this without directly modifying the material-ui library.


sastels (approved):
🦑

PR #272, opened 2018-06-05
Added Analytics to filter button presses

maxneuvians:
When you click a filter button, it gets logged in analytics

### DIFF
--- a/utils/analytics.js
+++ b/utils/analytics.js
@@ -1,7 +1,7 @@
 import ReactGA from "react-ga";
 
 export const initGA = () => {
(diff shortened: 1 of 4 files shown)


utils/analytics.js
```diff
@@ -1,17 +1,17 @@
 import ReactGA from "react-ga";
 
 export const initGA = () => {
-  ReactGA.initialize(process.env.GA_UA);
+  ReactGA.initialize("UA-102484926-4");
```
sastels: Why not get this from an env var?

sastels (approved):
minor question, 🦑

PR #275, opened 2018-06-05
Axe

maxneuvians:
Using axe-jest we can do some very basic accessibility testing in our app. Take note of the added code in __tests__.


sastels (approved):
cool!
And impressive that you found and fixed some accessibility bugs.

PR #281, opened 2018-06-05
Axe fix

maxneuvians:
Fixes a test that failed on master with the introduction of new axe testing

### DIFF
--- a/components/needs_selector.js
+++ b/components/needs_selector.js
@@ -93,7 +93,7 @@ NeedsSelector.propTypes = {
   selectedNeeds: PropTypes.object,
   t: PropTypes.func,
   theme: PropTypes.object,
-  clearNeeds: PropTypes.object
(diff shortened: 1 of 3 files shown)


components/needs_selector.js
```diff
@@ -93,7 +93,7 @@ NeedsSelector.propTypes = {
   selectedNeeds: PropTypes.object,
   t: PropTypes.func,
   theme: PropTypes.object,
-  clearNeeds: PropTypes.object
+  clearNeeds: PropTypes.func
```
smcmurtry, line 96: I introduced this bug. Thanks for catching that!

smcmurtry (approved):

components/needs_selector.js
maxneuvians: I should have merged master first again and I think we should have been good

PR #293, opened 2018-06-06
Who can get child benefits

### LINKED ISSUE #289: Show who can get child benefits
Ie something like

With this benefit, the Veteran may also have access to:
...
The family of the Veteran may also have access to:
...

sastels:
resolves #289 

Split list of child benefits into ones for the Veteran and ones for the family


smcmurtry (approved):
LGTM

PR #323, opened 2018-06-08
added language setting to the url

### LINKED ISSUE #297: Put the language in the url
when the switch language button is pressed we should add the `lng` to the url.

smcmurtry:
resolves #297

### DIFF
--- a/pages/A.js
+++ b/pages/A.js
@@ -116,14 +116,14 @@ export class A extends Component {
     this.setState({ width: window.innerWidth });
   }
 
-  setURL = () => {
-    let href = "/A?section=" + this.state.section;
-    if (Object.keys(this.state.selectedNeeds).length > 0) {
-      href += "&selectedNeeds=" + Object.keys(this.state.selectedNeeds).join();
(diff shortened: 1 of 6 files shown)


pages/A.js
```diff
@@ -116,24 +116,25 @@ export class A extends Component {
...
+    href += "&lng=" + this.props.t("current-language-code");
     Router.push(href);
   };
 
   setSection = section => {
     let newState = this.state;
     newState.section = section;
     this.setState(newState);
-    this.setURL(newState);
+    this.setURL();
```
sastels: I don't think this is safe. From what I've read, after calling `this.setState` it is not guaranteed that the state will be set before the next instruction (ie setState works asynchronously). So if the url is set before `setState` finishes, then the url will be set incorrectly and not fixed when `setState` finally finishes running. This is why I coded it here in this rather cumbersome way.

smcmurtry: Oh okay. I'll change it back

sastels (approved):
🦑

PR #325, opened 2018-06-08
put header and footer in Containers

### LINKED ISSUE #317: Enforce a max page width for header and footer
Currently on wide screens, Gov of Canada logos, buttons and feedback text all float to extreme right and left of window. Instead it should look as follows: 
screen shot 2018-06-07 at 5 06 27 pm

sastels:
resolves #317 

need to remove padding on alpha button (from gcui) though


smcmurtry (approved):
looks better

PR #326, opened 2018-06-10
fix close all button

### LINKED ISSUE #308: Close All button doesn't always work
It usually does? but sometimes it doesn't :(

sastels:
fixes #308  (maybe?)

dunno of this does it or not, we are getting a "calling setState on undefined object" error

### DIFF
--- a/components/BB.js
+++ b/components/BB.js
@@ -56,9 +56,11 @@ export class BB extends Component {
   collapseAllBenefits = () => {
     this.children.forEach(c => {
       c.setState({ open: false });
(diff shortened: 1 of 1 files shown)


maxneuvians (approved):
Tested for about 5 mins - did not seem to break. 🦑

sastels:
actually in hindsight it makes more sense to test for undefined right before doing a `setState`, if that is indeed the issue. (so now we are testing more cautiously, i.e. if we detected a bad state with my first commit, we will detect the same situation now and do even less).

Issue #320, opened 2018-06-08
make text on cards selectable


sastels:
can't seem to do it if they're clickable.

danprime:
While this is not specifically related to Material-UI there's an example in there where the text is still selectable
http://www.mit.edu/~rjc/aria/expandingSections.html#

PR #327, opened 2018-06-10
Make ExpansionSummary headings selectable

maxneuvians:
Address #320. When a div is fully clickable, the action can override the select event. One way to circumvent that is to use the CSS `user-select` property. (https://css-tricks.com/almanac/properties/u/user-select/). This allows a select event to select the entire text at once vs. being interrupted by the click event. The behavior is slight odd but it works. If you click on a title it will open the card and select the title, if you click somewhere outside of the card header it deselects it. To create the ideal situation of just making it selectable you would need to modify the ExpansionPanel component itself which might be more work that it is worth.

I also looked into testing if we can make sure that certain CSS rules are present on an element. Enzyme seems to only allow that for classes vs. individual rules.

### DIFF
--- a/components/benefit_cards.js
+++ b/components/benefit_cards.js
@@ -18,7 +18,11 @@ const styles = () => ({
   cardDescriptionText: {
     fontSize: "20px",
     fontWeight: 400,
-    padding: "15px 0px"
+    padding: "15px 0px",
+    "-webkit-user-select": "all",
+    "-moz-user-select": "all",
+    "-ms-user-select": "all",
+    "user-select": "all"
   },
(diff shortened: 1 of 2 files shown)


sastels:
Whoa, this is cool! I should learn more CSS. I think, though, that we want the title selectable and not the description? I’ll check with Michelle and Raluca tomorrow.

maxneuvians:
An alternative is to make only the buttons toggle the expansion. Might not be as intuitive.

maxneuvians:
Let us discuss this today before the PR becomes stale and we have too many merge conflicts

Issue #315, opened 2018-06-07
Say "All Benefits" when no filters selected


sastels:
instead of the default "Showing N benefits for you to consider"

PR #329, opened 2018-06-10
Added a method to count select options 

maxneuvians:
Shows "All benefits for you to consider" when there is no selection. Addresses #315

### DIFF
--- a/__tests__/components/BB_test.js
+++ b/__tests__/components/BB_test.js
@@ -203,4 +203,71 @@ describe("BB", () => {
   it("has a benefits counter", () => {
     expect(mounted_BB().find(".BenefitsCounter"));
   });
+
+  it("shows B3.All benefits to consider when no filters are selected", () => {
+    mounted_BB().setProps({
+      selectedEligibility: {
+        patronType: "",
+        serviceType: "",
+        statusAndVitals: ""
+      }
+    });
+    mounted_BB().setProps({ selectedNeeds: {} });
+    expect(
+      mounted_BB()
+        .find(".BenefitsCounter")
+        .last()
+        .text()
+    ).toEqual("B3.All benefits to consider");
+  });
+
+  describe("countSelection", () => {
+    it("returns 0 if nothing is selected", () => {
+      expect(
+        mounted_BB()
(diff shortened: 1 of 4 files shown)


components/BB.js
```diff
@@ -175,11 +175,22 @@ export class BB extends Component {
     this.setState({ sortByValue: event.target.value });
   };
 
+  countSelection = () => {
+    const reducer = (acc, obj) => acc + (Object.values(obj)[0] == null ? 0 : 1);
+    let count = Object.values(this.props.selectedEligibility).reduce(
```
sastels, line 180: Neat!

__tests__/components/BB_test.js
```diff
@@ -203,4 +203,71 @@ describe("BB", () => {
...
+        selectedEligibility: {
+          patronType: elegibilityPathsFixture[0].patronType
+        }
+      });
+      expect(
+        mounted_BB()
+          .instance()
+          .countSelection()
+      ).toEqual(2);
+    });
```
sastels, line 271: You’ve drunk the testing Koolaid I see. Nice.

components/BB.js
```diff
@@ -175,11 +175,22 @@ export class BB extends Component {
...
+      reducer,
+      0
+    );
+    return count + Object.values(this.props.selectedNeeds).length;
+  };
+
   countString = (x, t) => {
-    switch (x) {
-      case 0:
+    switch (true) {
```
sastels, line 188: Seems like a strange way to use a switch statement (Ie vs an if-else). Is the a common thing to do in JavaScript?

sastels (approved):
Minor comment, LGTM

__tests__/components/BB_test.js
maxneuvians: I figure being remote the tests need to be good :)

components/BB.js
maxneuvians: No JS - but in my mind makes it more readable. Happy to discuss.

sastels: nope, leave it as it is, I don't have strong opinions either way. 👍

Issue #318, opened 2018-06-07
Style/layout changes for top of page


smcmurtry:
<img width="1100" alt="screen shot 2018-06-07 at 5 06 27 pm" src="https://user-images.githubusercontent.com/5498428/41126330-28eadb7e-6a75-11e8-87c2-81ea52e17425.png">

sastels:
include making the background grey

PR #332, opened 2018-06-11
Layout changes

smcmurtry:
working on issue #318


sastels:
We could probably make the title/subtitle have a white background if we put it in its own <div> or something.

sastels (approved):
minor comment. LGTM

Issue #286, opened 2018-06-06
Change profile selector to be more reactive


sastels:
- initially just show patron type filter
- when that's answered (if not organization), show service type
- then show service status (don't show deceased if patron type is service person)

smcmurtry:
Based on today's discussion w/ design: not applicable options will be disabled (greyed out), not hidden

sastels:
Though I think the idea is for if nothing is selected then only the top filter group is displayed? In any case, we should start with a PR that just does the greying-out part, then if desired do a PR that hide things initially.

smcmurtry:
- [x] disable not-applicable options
- [x] make filter questions appear one at a time

sastels:
@maxneuvians The idea for the second part is that initially only the Recipient filter is shown, then when the user selects one, the Service Type is shown, then finally the Service Status. Also, if "Organizations" is selected then the other two filters are not shown.

PR #339, opened 2018-06-12
Added progressive opening for selectors with exception of organization

maxneuvians:
Addresses #286


sastels (approved):
awesome job!

sastels:
opps so if you select Vet, CAF, then Organization then the service status filter is still shown

PR #345, opened 2018-06-13
Split out BenefitList component from BB

### LINKED ISSUE #292: Make a benefit list component
re #180 
Extract the code in BB around the list of benefits, including text and sort / collapse buttons.

smcmurtry:
closes #292


sastels:
Something is funky with the Sort By dropdown - sometimes it works and sometimes doesn't. Seems to have to select the option twice to get it to resort.

maxneuvians (approved):
LGTM 🦑

PR #346, opened 2018-06-13
WIP: Search Basic implementation

maxneuvians:
The thing most confusing at this point is the stemming. It will match ill, but not illn, but then illnes. It looks like the stemming does it in groups of three. The expected behaviour is something more like this: 

https://bestbuyapis.github.io/api-documentation/ where even a one letter match produces a highlight.


maxneuvians:
Fixed the stemming issue, I will keep these tests red until I get some feedback on the feature

smcmurtry:
Wow, very impressive!

PR #351, opened 2018-06-14
Lunr search

maxneuvians:
New PR to merge the benefits list changes and to make it work with existing test


sastels:
So you could add it in and use a feature switch in the url to expose it, something like we do to load data from fixtures rather than airtable (for cypress) here
https://github.com/cds-snc/vac-benefits-directory/blob/master/pages/A.js#L103

maxneuvians:
Good idea, i added a `show_search` query param that will toggle the text box. Added tests all around for everything as well.

sastels (approved):
sweet! Awesome work!

Issue #353, opened 2018-06-14
Add progress bar to guided experience questions


smcmurtry:
Should be at bottom of page and have "% completed". 
https://app.zeplin.io/project/5acbd659dc553ec84805afd4/screen/5b232ed00df8d0e9691d2ff6

PR #355, opened 2018-06-14
Got the progress bar working more or less

smcmurtry:
Working on #353.

I created a new component so I wouldn't have to add the progress bar to GuidedExperienceProfile and GuidedExperienceNeeds. 

Any thoughts?

### DIFF
--- a/__tests__/components/guided_experience_needs_test.js
+++ b/__tests__/components/guided_experience_needs_test.js
@@ -39,7 +39,6 @@ describe("GuidedExperienceNeeds", () => {
       needs: needsFixture,
       selectedNeeds: {},
       setSelectedNeeds: jest.fn(),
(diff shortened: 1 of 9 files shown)


sastels:
I really like the idea you have here of having a container that contains either the profile ones, so we don't have to replicate the progress bar and other ui elements. I think it would be cleaner if the Profile or Needs components were passed as a child rather than passing a switch, ie so you do something like
```
<GuidedExperience  props >
  <GuidedExperienceProfile props />
</GuidedExperience>
```
This way we could separate out the props a bit. thoughts?

smcmurtry:
@sastels Great suggestion, I'll give it a try!

sastels (approved):
Awesome!!

Issue #331, opened 2018-06-11
Landing page


sastels:
Right now just a bit of text and links to the guided experience and the current app. ie something like

<img width="852" alt="image" src="https://user-images.githubusercontent.com/8228248/41470500-c4149f36-707e-11e8-8086-4541e130f874.png">

PR #359, opened 2018-06-15
Changed up landing page

maxneuvians:
Should look something like #331


smcmurtry (approved):
Looks good! Only issue is that the background colour doesn't take up the full screen which looks a bit weird on a wide screen.

maxneuvians:
I figure this is a bigger layouting issue with the fixed with container.

PR #376, opened 2018-06-18
adjust GE options for WSV

### LINKED ISSUE #369: Guided Experience: Selecting veteran + WWII should hide the still-serving option
This logic is already implemented in the RadioSelector component. Is it possible to re-use this?

sastels:
fixes #369 

for WSV:
* hide "still serving" option
* if patronType === Veteran, skip statusAndVitals question entirely (only 1 valid option, "released")

### DIFF
--- a/locales/en/common.json
+++ b/locales/en/common.json
@@ -114,8 +114,6 @@
   "service-person": "Veteran or serving member",
   "family": "Family members",
   "organization": "Organizations",
-  "released": "Released",
-  "still serving": "Still Serving",
   "deceased": "Deceased",
   "disabled-serviceStatusFilter": "Only required for CAF Veterans",
   "disabled-servicePersonVitalStatusFilter": "Only required for family",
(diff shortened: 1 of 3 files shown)


maxneuvians:
Two small issues:

* Click Veterans or Serving
* Click WW2
= Back button is broken

* Click Veterans or Serving
* Click CAF
* Click Still serving
* Click back
* Click WW2
= Can no longer change Still Serving

Both are not big problems, especially the second one is very conditional.

sastels:
👍  good catch!

maxneuvians (approved):
🦑 LGTM

locales/en/common.json
```diff
@@ -114,8 +114,6 @@
   "service-person": "Veteran or serving member",
   "family": "Family members",
   "organization": "Organizations",
-  "released": "Released",
-  "still serving": "Still Serving",
```
smcmurtry, line 118: Thanks for cleaning up some of the unused variables - I'm sure there are more. We should probably make an issue to clean it up and put everything in sections.

smcmurtry (approved):
Looks good!

locales/en/common.json
sastels: Probably, though better might be to move most of this text to AirTable (though we'd still have to figure out what's still needed). putting things in sections would probably help, mind you.

PR #378, opened 2018-06-18
Hide statusAndVitals filter for Veteran / WSV

### LINKED ISSUE #377: Hide statusAndVitals filter for Veteran / WSV in BB
there's only one valid option, "Released", so there's no point in showing the filter.

sastels:
fixes #377

In BB if the profile is Veteran / WSV, then there's only one legal statusAndVitals choice, "Released".
* auto-select "Released" (in case there's some benefits that require Released status)
* hide filter

### DIFF
--- a/components/BB.js
+++ b/components/BB.js
@@ -245,6 +245,15 @@ export class BB extends Component {
 
   render() {
     const { t, classes } = this.props; // eslint-disable-line no-unused-vars
+
+    if (
+      this.props.selectedEligibility.patronType === "service-person" &&
+      this.props.selectedEligibility.serviceType === "WSV (WWII or Korea)" &&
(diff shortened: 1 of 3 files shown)


sastels:
Also clear "still serving" if user chooses WSV.

smcmurtry:
One small issue: selecting:
veteran
WWII
CAF

Then the status question shows released selected, where it should be none selected.

sastels:
yeah, so the reason for that is because selecting Vet / WWII (now) causes Released to be autoselected. I couldn't see a good way to clear that released if you move off WWII. It is confusing though. I think the correct approach is to not auto-select Released, and if there are benefits that require released, they should have an eligibility path to Vet / WSV / na .

maxneuvians (approved):
🦑 LGTM

PR #382, opened 2018-06-18
Bookmarking

sastels:
Adds ability to "favourite" benefits.


sastels:
I think I need to hook up server side rendering to the cookies, as in
https://www.npmjs.com/package/universal-cookie

components/benefit_cards.js
```diff
@@ -1,18 +1,21 @@
...
 import ExpansionPanelSummary from "material-ui/ExpansionPanel/ExpansionPanelSummary";
 import ExpansionPanelDetails from "material-ui/ExpansionPanel/ExpansionPanelDetails";
+import Favorite from "@material-ui/icons/Favorite";
+import FavoriteBorder from "@material-ui/icons/FavoriteBorder";
 import Highlighter from "react-highlight-words";
 
 import { logEvent } from "../utils/analytics";
 
 const styles = () => ({
+  iconButton: {},
```
smcmurtry: 🔥

components/benefit_cards.js
```diff
@@ -125,6 +137,22 @@ export class BenefitCard extends Component {
                         : benefit.vacNameFr
                     }
                   />
+                  {this.props.showFavourite ? (
+                    <IconButton
+                      className={classes.iconButton}
```
smcmurtry: 🔥

components/benefit_cards.js
```diff
@@ -125,6 +137,22 @@ export class BenefitCard extends Component {
...
+                      aria-label="Favorite Button"
```
smcmurtry: nice!

components/BB.js
```diff
@@ -384,6 +384,14 @@ export class BB extends Component {
                 </Grid>
 
                 <Grid item xs={6} className={classnames(classes.collapse)}>
+                  <Button
+                    id="Favourites"
+                    variant="flat"
+                    size="small"
+                    onClick={() => this.props.setSection("favourites")}
+                  >
+                    {t("B3.favouritesButtonText")}
```
smcmurtry, line 391: I like this - we should probably be doing this more. Say what the text is for rather than what the text is since it can change and that gets confusing.

smcmurtry:
I just tried out the search feature on the favourites section - the behaviour is different than search on BB. Entering a search term reduces the list of benefits in BB but doesn't reduce the list in favourites.

Do we need search in favourites? Maybe it would make sense to leave it out for now.

smcmurtry:
Otherwise this looks really good!

maxneuvians (approved):
LGTM2 🦑

Issue #367, opened 2018-06-15
Skipping GE questions produces some funny results


maxneuvians:
Start GE
Click Next
Click Next
Click Deceased

Nothing is selected on result page.

Should next only be clickable if something was selected?

smcmurtry:
I just chatted with Michelle - she suggests that if you click Next from the first question with nothing selected, it skips to the Needs question

maxneuvians:
Closed in #394

PR #394, opened 2018-06-19
GE go to Needs if a question is skipped

### LINKED ISSUE #387: GE skip to needs if profile question not answered
otherwise BB will hide responses :(

sastels:
fixes #387 and #367

in GE, go directly to Needs if a question is skipped.


smcmurtry (approved):
Looks good to me! Maybe check with design before you merge since they're using the azure link for user testing today.

sastels:
I sent the Heroku link to Michelle to use for testing but will double check just in case 👍

PR #407, opened 2018-06-20
Printing

### LINKED ISSUE #276: Printing
Ability to print what they have selected (filters & needs) and the filtered benefits. Instructions included (ex. 1800 number)

https://github.com/gregnb/react-to-print

smcmurtry:
resolves #276 

A bit rough but it works.

Benefits list component needs to be replaced with something that will display the expanded content of the cards with minimal formatting for printing.

### DIFF
--- a/__tests__/pages/print_test.js
+++ b/__tests__/pages/print_test.js
@@ -57,7 +57,7 @@ describe("Print", () => {
 
   it("parses url correctly", () => {
     props.url.query["patronType"] = "service-person";
-    props.url.query["serviceType"] = "WSV%20(WWII%20or%20Korea)";
+    props.url.query["serviceType"] = "WSV (WWII or Korea)";
     props.url.query["statusAndVitals"] = "releasedAlive";
     props.url.query["needs"] = "43534534,43534ewr534";
     expect(
--- a/__tests__/pages/print_test.js
+++ b/__tests__/pages/print_test.js
@@ -1,0 +1,75 @@
+/* eslint-env jest */
+
+import { shallow } from "enzyme";
+import React from "react";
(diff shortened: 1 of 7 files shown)


components/BB.js
```diff
@@ -254,8 +283,15 @@ export class BB extends Component {
       this.props.selectedNeeds
     );
 
+    const url = this.getPrintUrl(
```
sastels: maybe name this printUrl or something, we have a lot of urls in this app.

sastels:
Slightly odd behavior when spitting across two pages. We can punt on that for now if you can’t see a quick fix
9c76d61c-59eb-40dd-8a35-87e3c0043eab

sastels:
Actually when we replace benefit list as you mention above we can ensure that it splits correctly. 👍

smcmurtry:
Haha, you really printed! Yes I think we should create a `PrintableBenefitsList` component.

__tests__/pages/print_test.js
```diff
@@ -0,0 +1,75 @@
...
+    expect(await axe(html)).toHaveNoViolations();
+  });
+
+  it("renders BenefitList component", () => {
+    expect(mountedPrint().find("BenefitList").length).toEqual(1);
+  });
+
+  it("parses url correctly", () => {
+    props.url.query["patronType"] = "service-person";
+    props.url.query["serviceType"] = "WSV%20(WWII%20or%20Korea)";
```
sastels: I think you could just put spaces here instead of `%20`? Otherwise wouldn't our app break? 🤔

sastels (approved):
This is awesome!

sastels:
(opps sorry for yet another merge conflict)

PR #411, opened 2018-06-21
Fix for selectability

### LINKED ISSUE #320: make text on cards selectable

maxneuvians:
Fixes #320


sastels (approved):
Wicked!! 🦑

smcmurtry:
This is wonderful! Can we apply it to the expansion panels that hold the radio buttons and needs too?

PR #421, opened 2018-06-21
Hide refresh cache button

smcmurtry:


__tests__/components/menu_bar_test.js
```diff
@@ -90,4 +90,13 @@ describe("MenuBar", () => {
       "other-language"
     );
   });
+
+  it("refresh cache button shown if showRefreshCache is true", () => {
```
sastels, line 94: What about a test that it’s hidden for showRefreshCache undefined or false?

sastels (approved):
Minor comment, LGTM

Issue #408, opened 2018-06-20
Refine printing


sastels:
- [ ] move the printing functions to the printing page and make them exportable
- [ ] actually, maybe make a print button component that we can give the eligibility, needs, and benefits?
- [x] add the print button to the favorites page
- [x] make the print button prettier
- [x] probably don’t need a PrintBenefitList component since we just need this in the printing component?
- [x] probably just unroll the list in the print page and manually include the desired fields (Ie instead of a benefit card)?

sastels:
We've done most of this, will refine further in collaboration with Design.

PR #423, opened 2018-06-22
Add print button to Favourites  page

sastels:
re #408

### DIFF
--- a/components/BB.js
+++ b/components/BB.js
@@ -368,6 +368,14 @@ export class BB extends Component {
 
                 <Grid item xs={6} className={classnames(classes.collapse)}>
                   <Button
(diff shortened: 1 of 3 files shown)


components/BB.js
```diff
@@ -368,21 +368,21 @@ export class BB extends Component {
 
                 <Grid item xs={6} className={classnames(classes.collapse)}>
                   <Button
+                    id="Favourites"
```
sastels: I just flipped the order of the buttons here, and GitHub got confused :(

smcmurtry (approved):
Looks good! If we were going to keep getPrintUrl we could make it an importable function in utils.

PR #425, opened 2018-06-22
Change page title of data validation page

sastels:
i.e. the text on the browser tab


smcmurtry (approved):
Nice! We should do this for all the pages.

PR #428, opened 2018-06-22
Data-Val-Needs

### LINKED ISSUE #401: Data Validation: benefits without a need
Add a test that ensures that all benefits have at least one associated need

apmcguigan:
resolves #401

### DIFF
--- a/__tests__/fixtures/benefits.js
+++ b/__tests__/fixtures/benefits.js
@@ -21,6 +21,7 @@ const benefitsFixture = [
     oneLineDescriptionEn: "English benefit Description",
     oneLineDescriptionFr: "Description de l'avantage français",
     availableIndependently: "Requires Gateway Benefit",
+    needs: ["3"],
     examples: ["0", "1"]
   },
   {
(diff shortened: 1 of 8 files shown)


apmcguigan:
Not sure what happened with the conflicting package.json. Was having some issues getting yarn test to run so had to mess with Node_Modules folder

sastels:
Did you do a fresh `git pull master` before creating the `Data-Val-All-benefits-have-needs` branch?

apmcguigan:
I'm not sure. I think I did one yesterday but then didn't work on the branch until earlier today.

apmcguigan:
Why was test removed from package.json? Is that something that is only needed for setup on my end?

maxneuvians:
I don't think it is something we need.

https://www.npmjs.com/package/test

maxneuvians (approved):
LGTM 🦑

sastels (approved):
👍 good job - and it found benefits we don't have needs for! :)

PR #438, opened 2018-06-26
Spin

### LINKED ISSUE #437: Make it obvious that results have changed when a new profile/need is selected
This is a request from design based on user feedback. 

Proposed solution is to reload the page and scroll so that "Showing x benefits for you to consider" is at the top of page, whenever a new profile/need is selected in BB.

maxneuvians:
~~Jump to top of page on Benefits list~~ change fix #437


sastels (approved):
🦑

PR #446, opened 2018-06-27
Checks Eligibility Path

### LINKED ISSUE #399: Data Validation: all benefits appear in an eligibility path
Add a test to ensure that every benefit is reachable through an eligibility path. (otherwise the app will ignore benefits)

apmcguigan:
Closes #399


apmcguigan:
Should yarn.lock be in gitignore?

sastels:
I think `yarn.lock` works with `yarn sync` so that in CI / wherever you can be confident that the versions of the modules being used are the same as those that were used locally when the code was being developed. @maxneuvians @smcmurtry Is this true?

related question: should we use `yarn sync` here
https://github.com/cds-snc/vac-benefits-directory/blob/master/.circleci/config.yml#L29

sastels:
Can you add a test (similar to the one that tests the "benefits without needs" line)?

__tests__/pages/data-validation_tests.js
```diff
@@ -87,4 +87,14 @@ describe("DataValidation", () => {
     props.benefits[0].benefitPageFr = "";
     expect(mountedDataValidation().html()).toContain("Fail");
   });
+
+  it("fails if a benefit does not have any linked Needs", () => {
+    props.benefits[0].benefitPageFr = "";
+    expect(mountedDataValidation().html()).toContain("Fail");
+  });
+
+  it("fails if a benefit is not connected to any Eligibility Paths", () => {
```
sastels: 👍

__tests__/pages/data-validation_tests.js
```diff
@@ -87,4 +87,14 @@ describe("DataValidation", () => {
     props.benefits[0].benefitPageFr = "";
     expect(mountedDataValidation().html()).toContain("Fail");
   });
+
+  it("fails if a benefit does not have any linked Needs", () => {
+    props.benefits[0].benefitPageFr = "";
```
sastels: Oh,, I though there was already a test for no linked needs - how'd I not complain about this last time? :)

Thought your test here is just deleting the French link rather than the needs column.

pages/data-validation.js
```diff
@@ -46,7 +46,8 @@ export class DataValidation extends Component {
   }
 
   checkEligibiltyPaths(b) {
-    return !(b.eligibilityPaths && b.eligibilityPaths != "");
+    console.log(b.elibigbilityPaths);
+    return !(b.elibigbilityPaths && b.elibigbilityPaths != "");
```
sastels: Wait, I thought I fixed that in AirTable? We should fix the AirTable typo and use the correct spelling here.

sastels:
ah, I merged in my PR deleting the common.json files 😁 You'll have to add the new strings to AirTable's "Text" table. It should just work then.

apmcguigan:
so I should either delete common.json or add it to gitignore? @sastels

sastels:
👍 , move the strings to AirTable and 🔥 the common.json files

sastels:
ugh, stupid Cypress errors. I think we should 🔥 our cypress tests until we figure if / how we want to do them. Let me put a PR in for that.

apmcguigan:
I could redownload master and just add these few changes in would probably be the easiest fix, because without your new airtable changes this is gonna fail forever

sastels (approved):
Looks good, tests should pass if you do a `git pull origin master`, then you should be good to go!

PR #448, opened 2018-06-27
I18n to airtable

### LINKED ISSUE #435: Move app text strings to AirTable
Move most if not all of the i18n strings to AirTable.
Change keys so they indicate what the text is used for rather than exact English words.

sastels:
resolves #435 

Move strings out of `locales/*/common.json` and into AirTable


cypress/integration/page_A_spec.js
```diff
@@ -6,9 +6,9 @@ describe("Page A", () => {
   });
 
   it("Has a language button that works", () => {
-    cy.contains("Veteran Affairs Canada");
-    cy.contains("Anciens Combattants Canada").should("not.exist");
-    cy.contains("Français").click();
-    cy.contains("Anciens Combattants Canada");
+    // cy.contains("Veteran Affairs Canada");
```
smcmurtry, line 9: Should we just kill this test? Or fix it?

cypress/integration/page_A_spec.js
```diff
@@ -6,9 +6,9 @@ describe("Page A", () => {
...
+    // cy.contains("Anciens Combattants Canada").should("not.exist");
+    // cy.contains("Français").click();
+    // cy.contains("Anciens Combattants Canada");
```
sastels, line 12: decided to just 🔥 this test for now. We have to sit down and decide if we want cypress testing, and if so we really should pretty much scrap all the existing testing and start from scratch.

smcmurtry (approved):

PR #453, opened 2018-06-28
Upgraded to next 6.1

maxneuvians:
Next 6.1 came out, seems like we can upgrade


sastels (approved):
seems to still work. 🚀

PR #456, opened 2018-06-28
Disable cypress

sastels:
Right now our Cypress tests are shit, there's no point in running them. Let's disable them for now, and if we want to do proper Cypress testing at a later date we can rewrite them and re-enable.


maxneuvians (approved):
🚒 🔥 🦑

PR #462, opened 2018-06-28
Get selectedNeeds from redux

### LINKED ISSUE #461: get selectedNeeds from redux rather than props

sastels:
resolves #461


maxneuvians (approved):
LGTM 🦑

PR #463, opened 2018-06-29
Fix print sorting

### LINKED ISSUE #459: printing has benefits in wrong order
when opening the print page, get an error
```
Warning: Failed prop type: The prop `sortByValue` is marked as required in `Print`, but its value is `undefined`.
```
Also (probably as a consequence) the benefits are not sorted in the correct order.

sastels:
fixes #459 

print page didn't have benefits sorted anymore. Fixed now.

I think the bug was introduced when we changed the print page from using BenefitList (which sorts the benefits internally) to manually unrolling the filteredBenefits.

Note that eventually we will probably move the sortedBenefits into redux and just pull that into the print page.

### DIFF
--- a/__tests__/components/BB_test.js
+++ b/__tests__/components/BB_test.js
@@ -299,7 +299,7 @@ describe("BB", () => {
       expect(mounted_BB().state().frIdx).not.toEqual(null);
     });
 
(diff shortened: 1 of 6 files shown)


maxneuvians (approved):
Does what the label says 🦑

PR #470, opened 2018-06-29
Display failing rows

### LINKED ISSUE #400: Data Validation: display failing rows
Add a section below the table that indicates which rows of which table are failing which test. ie something like
```
<failing test name>:
benefits: id4w593485, id2435820354, id23457923, id43534435
...
```

apmcguigan:
resolves #400


maxneuvians (commented):
Awesome work, we should probably have a test for `getBrokenBenefits`. There is some logic in there and our test suite fails if we don't cover at least 75% of lines.

sastels (approved):
awesome! 🦑

Issue #477, opened 2018-07-03
Circle CI Build Warning


maxneuvians:
```
PASS  __tests__/components/profile_selector_test.js
(node:218) UnhandledPromiseRejectionWarning: Error: only absolute urls are supported
(node:218) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). (rejection id: 1)
(node:218) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code
```

Should probably look into this.

maxneuvians:
Tracked to feedback_bar:

```
$ /home/circleci/repo/node_modules/.bin/jest components/feedback_bar_test.js
(node:3171) UnhandledPromiseRejectionWarning: Error: only absolute urls are supported
(node:3171) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). (rejection id: 1)
(node:3171) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
 PASS  __tests__/components/feedback_bar_test.js
```

PR #478, opened 2018-07-03
Fixed Test warning by adding a catch

maxneuvians:
Feedbackbar component was throwing a warning in testing #477


sastels (approved):
seems good.

PR #482, opened 2018-07-03
Make BB its own page

### LINKED ISSUE #442: Make BB its own page
After #443 is complete, excise BB from A and put on its own page

sastels:
resolves #442 

Make BB the `benefits-directory`page, hooked into `A`

### DIFF
--- a/__tests__/components/guided_experience_test.js
+++ b/__tests__/components/guided_experience_test.js
@@ -33,6 +33,7 @@ describe("GuidedExperience", () => {
       classes: {},
       id: "YY",
       nextSection: "ZZ",
+      benefitsDirectoryUrl: "/benefits-directory",
       prevSection: "XX",
       stepNumber: 1,
       children: <div className="thing" />,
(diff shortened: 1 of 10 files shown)


__tests__/components/guided_experience_test.js
```diff
@@ -33,6 +33,7 @@ describe("GuidedExperience", () => {
       classes: {},
       id: "YY",
       nextSection: "ZZ",
+      benefitsDirectoryUrl: "/benefits-directory",
```
maxneuvians, line 36: At some point we should talk about making a Routes file just with the URLs as constants, so when the URL changes because of preference, we do not need to update the code. Somewhat similar to translation strings.

maxneuvians (approved):
🦑 Really solid work!

__tests__/components/guided_experience_test.js
sastels: 👍 yeah, I was thinking something along those lines.

maxneuvians:
When you merge please also close #180 - we did it!

Issue #406, opened 2018-06-20
Revise footer feedback


sastels:
* remove right footer question (pops up if "No" selected)
* revise yes/no question text (text to be provided by design)

maxneuvians:
~remove right footer question (pops up if "No" selected)~ in #483

sastels:
Style as in
 https://app.zeplin.io/project/5acbd659dc553ec84805afd4/screen/5b16f12431e8b89541f8cf09

![image](https://user-images.githubusercontent.com/8228248/43470882-a6126484-94b7-11e8-9288-c9d7c97bc1e3.png)

PR #483, opened 2018-07-03
Removed comment prompt

maxneuvians:
Addresses:

`remove right footer question (pops up if "No" selected)`

in #406


sastels (approved):
LGTM. Though kinda strange that once the box pops open you have to scroll down since the form is not on the screen. Do you think it would be better if clicking "No" opened the dialog and moved you to the bottom of the page?

maxneuvians:
Agreed! It should focus on the first text field. I will include that in the PR.

Issue #356, opened 2018-06-15
Prepopulate searchString from query param


maxneuvians:
the search bar should  be pre-populated from a query string and also add to the query string when it changes. This will allow a search filtered result to be bookmarked

maxneuvians:
Closed in #494

PR #494, opened 2018-07-04
Adds searchQuery to url

maxneuvians:
searchQuery has an impact on state and should be reflected in the URL. #356


sastels:
Is it easy to set the search text if the url is set?
image

maxneuvians:
Yes! Good catch!

sastels (approved):
🦑

PR #503, opened 2018-07-05
rename airtable text to translations

### LINKED ISSUE #501: rename AirTable "text" table to "translations"
makes more sense.

sastels:
fixes #501 

for now I actually copies the AirTable table so we wouldn't break everything (including the current user testing branch). Will delete the old column once the new code is merged / used everywhere.


utils/airtable.js
```diff
@@ -1,31 +0,0 @@
-import fetch from "isomorphic-unfetch";
```
sastels, line 1: Also deleted this file as we don't use it anymore.

apmcguigan (approved):
LGTM

PR #511, opened 2018-07-06
Tweak needs tags

### LINKED ISSUE #510: tweak needs tags
from discussions with Design, change to rectangular white text on black background.

sastels:
resolves #510 

change the needs tags to look more like the selector buttons. (and also less like buttons)


smcmurtry (approved):
Looks good!

PR #519, opened 2018-07-06
Fix Heroku R10 timeouts

### LINKED ISSUE #349: Memory usage
Apps on Heroku are occasionally exceeding memory and failing (currently a 512MB limit). Can we lower the amount of memory the server needs?

sastels:
I think this might fix #349 (R10 errors causing a time out on Heroku).
Following 
https://help.heroku.com/P1AVPANS/why-is-my-node-js-app-crashing-with-an-r10-error
and
https://devcenter.heroku.com/articles/getting-started-with-nodejs#define-a-procfile
I took the `yarn build` out of the `Procfile` and added it as a postinstall script in `packages.json`

### DIFF
--- a/Procfile
+++ b/Procfile
@@ -1,1 +1,1 @@
-web: yarn build && yarn start
+web: yarn start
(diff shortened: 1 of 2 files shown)


maxneuvians (approved):
Looks like the PR app built - so let us give it a try 🦑

PR #528, opened 2018-07-10
Option B filter logic

### LINKED ISSUE #520: List of benefits should only show benefits that match search / filter criteria
Should be behind feature flag (see #502)

sastels:
resolves #520 

Make a new version of the selector that filters benefits based on selected profile and needs. In this version, pass along any benefits that match both (regardless of parent/child relationships or independent status).


maxneuvians (approved):
LGTM 🦑

PR #529, opened 2018-07-10
Move needs tags to bottom

### LINKED ISSUE #524: Need tags should be on a new line

sastels:
resolves #524 

https://vac-poc-staging-pr-529.herokuapp.com/benefits-directory?lng=en&option=B


smcmurtry (approved):
Nice! We may want to move to a `Grid` layout for the card summary eventually.

PR #531, opened 2018-07-10
Fixed console warning by properly mocking

maxneuvians:


sastels (approved):
that's some good moching.

Issue #539, opened 2018-07-11
Reorder layout of buttons (bookmark, print, search) on benefits listing


maxneuvians:
<img width="637" alt="screen shot 2018-07-11 at 8 29 49 am" src="https://user-images.githubusercontent.com/867334/42571286-ad6d6dc4-84e4-11e8-8343-169a7c83d057.png">

PR #540, opened 2018-07-11
Moved buttons around in BB

maxneuvians:
Moved buttons according to design #539


smcmurtry (approved):
Looks great! 🦑

maxneuvians:
Closes #539

PR #541, opened 2018-07-11
Updated header colors and added div to hold gradient

thomascraig:


sastels:
@thomascraig do we still want this PR?

sastels:
We're probably not going to merge this PR, but let's keep the branch around for now in case we want to look at applying those styling changes.

PR #561, opened 2018-07-12
Card headers

### LINKED ISSUE #497: Benefit card should have header if it has a parent
screen shot 2018-07-05 at 7 10 13 am

sastels:
resolves #497 

Add a header to option B cards if it has a parent.


maxneuvians (approved):
Nice work! 🦑

PR #575, opened 2018-07-16
Changed default to card B

### LINKED ISSUE #574: Default to option B
If no option flag is set, default to B not A. Option B is what design wants to user test, so this will make testing simpler.

apmcguigan:
set option=A to get old cards
closes #574


sastels (approved):
LGTM! 🦑

PR #588, opened 2018-07-17
Add Snyk badge to Readme

sastels:
Tracker does so why shouldn't we?


maxneuvians (approved):
Looks suspicious, but I will agree to this based on the mutual trust we have built over the years 🦑

PR #620, opened 2018-07-18
Make radio buttons blue when selected

### LINKED ISSUE #609: change radio button color
change from pink to the blue we're using.

image

sastels:
resolves #609 

change selected radio buttons in guided experience and benefits directory from pink to blue.


maxneuvians (approved):
🦑 LGTM, does not even flicker

PR #627, opened 2018-07-19
fix skipped question in GE

### LINKED ISSUE #622: Family / WSV skips to health question
skips the status released / deceased question

sastels:
fixes #622 

in GE, family / WSV skipped the released / still serving question.


maxneuvians (approved):
🦑 LGTM

Issue #613, opened 2018-07-18
Clear health question


maxneuvians:
https://cdsvacpoc.azurewebsites.net/benefits-directory?lng=en&patronType=service-person&serviceType=WSV%20(WWII%20or%20Korea)&serviceHealthIssue=false

Click Service Type: RCMP
Click Service Status: Released
Health question is still selected.

PR #638, opened 2018-07-20
[WIP] Clear filter when hidden

smcmurtry:
Not sure how to fix this.


smcmurtry:
I decided to do this the straight forward way: #613

Maybe we can revisit this later

PR #641, opened 2018-07-23
fix minor version of docker node in ci

sastels:
partial work on #642 
master failing on ci due to change to docker node image being used.


maxneuvians (approved):
LGTM 🦑

PR #643, opened 2018-07-23
Remove jump to top of page when clicking need on mobile

### LINKED ISSUE #635: On mobile, clicking a need (or a profile Q) should not jump to top of page
Request from Michelle. 

On wider screens this is working as intended, but on narrow screens you can't actually see the benefits at the top of the screen.

smcmurtry:
Closes #635


sastels (approved):
LGTM, let's get master fixed first though.

Issue #564, opened 2018-07-13
Fix black header bar on mobile


sastels:
Might want to change the language button options to "Fr" and "En", but even with that the flag logo would have to shrink a bit.

![image](https://user-images.githubusercontent.com/8228248/42699102-bff4558e-868d-11e8-99f3-6960dbd88660.png)

maxneuvians:
Fixed in #687

PR #687, opened 2018-07-25
Fix for #564

maxneuvians:
Set federal logo to dynamic width - this should solve some of the problems with #564. Tested it on Android and it looked fine. Probably a lot of the other recent mobile layout fixes helped this as well.

### DIFF
--- a/components/federal_banner.js
+++ b/components/federal_banner.js
@@ -61,7 +61,7 @@ class FederalBanner extends Component {
       <div className={container}>
         <div className="svg-container">
           <GoCSignature
-            width="320px"
+            width="100%"
             lang={t("current-language-code")}
             flag="#fff"
             text="#fff"


components/federal_banner.js
```diff
@@ -61,7 +61,7 @@ class FederalBanner extends Component {
       <div className={container}>
         <div className="svg-container">
           <GoCSignature
-            width="320px"
+            width="100%"
```
sastels, line 64: well that was an easy fix!

sastels (approved):
LGTM. We should make a list of devices "wide ones", "narrow ones", etc and check layout things on them.

maxneuvians:
Hillary is setting up a device lab, but I know chrome has a plugin that will size your screen accordingly.

Issue #691, opened 2018-07-25
Search: Pressing enter completes search


maxneuvians:
Can we start the search when the user hits the return key in the search bar (main page) rather than having to use the search button

Issue #690, opened 2018-07-25
Search: Use lunr on main page


maxneuvians:
On main page, typing "benefits" stops showing "caregiver benefit" as an option. Is there any way to do this?

PR #704, opened 2018-07-25
Lunr upgrades

maxneuvians:
Started as a fix for #690 and #691 but also fixed a bug with not ORing words together in search


sastels:
If I put benefits on the main page all the drop down suggestions go away when I type the 's'

sastels:
And nothing appears in the list on BB

maxneuvians:
I lowercased and hopefully it works - should you not be on a flight @sastels?

sastels (approved):
LGTM 🦑 
(Flight delayed 30 min)

PR #724, opened 2018-07-30
Fix for #718

### LINKED ISSUE #718: Longer load time for Saved Benefits
To replicate:

1. Visit the directory
1. Mark a benefit for save for later
1. Click on 'View saved benefits'
<------ Notice that it takes around 3 seconds before the save benefit is displayed.

The noticeable delay is between when the page defaults to (0) Save Benefits and the time it shows the loading icon and then the actual benefits.

I profiled it in Google Inspector (and I saved the profiling data but Github doesn't allow me to submit it).

screen shot 2018-07-27 at 11 23 28 am

maxneuvians:
Removed the component state based favourites loading as we were already using redux for both reading and writing. Also fixes #718


smcmurtry (approved):
Well done! Nice when you can remove 40 lines of code and the app works better 😄

Issue #738, opened 2018-07-31
Button focus state issue - a11y


melodyma:
The pulsing focus states for buttons can make someone feel nauseous/trigger motion sickness, especially for the clientele you're servicing. I would suggest to stick with simple focus states.

maxneuvians:
Fixed in #754

melodyma:
Hi @maxneuvians - Thanks for fixing the ripple animation for focus on mouse so quickly! However, it's not fixed for keyboard tab. Opened new issue #796

PR #754, opened 2018-08-01
Remove ripple focus animation and also increased border size for focu…

### LINKED ISSUE #737: No focus states for Categories - a11y issue
When using keyboard tabbing into Categories, cannot tell that you're focusing on any categories because there are no focus states for them.

maxneuvians:
…s categories

Fixes #737 and #738


sastels (approved):
nice!

PR #758, opened 2018-08-01
Print benefits splitting

### LINKED ISSUE #746: Stop benefits from being split on page break
screen shot 2018-08-01 at 10 15 40 am

### LINKED ISSUE #747: Incorrect font on print page
screen shot 2018-08-01 at 10 19 10 am

smcmurtry:
Closes #746 
Closes #747 

Finally figured it out - the issue was I was putting the table inside a `Grid` element, which was causing everything inside it to be `div`s.

### DIFF
--- a/pages/print.js
+++ b/pages/print.js
@@ -48,7 +48,7 @@ const profile_questions = [
 
 export class Print extends Component {
   componentDidMount() {
-    // window.print();
+    window.print();
   }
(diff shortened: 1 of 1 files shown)


maxneuvians (approved):
Tested and works for me. Found another issue, but not related to PR 🦑

PR #761, opened 2018-08-01
Added home link to federal logo

### LINKED ISSUE #645: Universal home button
We need a way to get back to the landing page from all pages

maxneuvians:
Fixes #645


sastels:
Is that a good place for it? Seems like clicking on "Government of Canada" should go to canada.ca or something.

maxneuvians:
We need to have it somewhere and eventually that will be replaced by a vac logo - it just has been a pain to always navigate back using the url or back button.

PR #771, opened 2018-08-02
search by address rather than lat/long

### LINKED ISSUE #770: get directions by address rather than lat/long
maps search by lat/long can lead to a different address ☹️

sastels:
fixes #770


sastels:
works on desktop (google maps) as well as ios (apple maps)

sastels:
noticed by @smcmurtry french addresses don't always work in the search 😞

maxneuvians (approved):
LGTM 🦑

PR #777, opened 2018-08-02
Send Github data only on stats

maxneuvians:
Only send github stats on stats page


smcmurtry (approved):
11.8mb to 4.6mb on /benefits-directory! Nice work.

PR #778, opened 2018-08-02
Fix back buttons

### LINKED ISSUE #720: Back Button won't work if you click a link.
If you were to use a link or copy paste directly to map/favourites the back button will bring you back to where you were before then. rather than the directory.

smcmurtry:
Closes #720


maxneuvians (approved):
Nice work! 🦑

PR #793, opened 2018-08-03
Janky filter dropdown

### LINKED ISSUE #790: stop filter expand/contract on mobile load
It's janky.

sastels:
fixes #790 
Use breakpoints to stop ProfileNeedsSelector flicker on mobile (where it would first appear as just a box, then switch to an open expansion panel, then close).  Decided that the cleanest way was to make two components, one for desktop and one for mobile. Had to futz with ids to get the axe tests to not complain about repeated ids.


smcmurtry (approved):
Looks good!

PR #827, opened 2018-08-07
Fix print button

### LINKED ISSUE #818: Print Button clears Needs

sastels:
fixes #818 

Got rid of the Link in Print button. This required adding the ids for the closest and selected area offices into the print page url.


__tests__/components/BB_test.js
```diff
@@ -200,7 +204,9 @@ describe("BB", () => {
         { patronType: "service-person" },
         { need1: "need1", need2: "need2" },
         "sorting",
-        "en"
+        "en",
+        "0",
+        "0"
       );
     expect(url).toEqual(
       "print?lng=en&patronType=service-person&needs=need1,need2&sortBy=sorting&benefits=id1,id2"
```
maxneuvians: How is this test passing? Should it not be `&closestAOID=0&selectedAOID=0`

sastels: two wrongs make a right? The `getPrintUrl` function actually expects the area office objects, not their ids as I was sending.

sastels: (btw, thanks - good catch!)

maxneuvians: The sad thing is that Javascript just went at head with "0" as undefined because it should have appended just the empty GET names.

sastels: actually the code happily computes `"0".id`, got undefined, went its merry way.

maxneuvians (approved):
Nice work! 🦑

PR #830, opened 2018-08-07
Updated feedback email in alpha banner

smcmurtry:
We now have two emails set up for product feedback!

vac@cds-snc.ca for english
and 
acc@cds-snc.ca for french


maxneuvians (approved):
🦑 LGTM

PR #832, opened 2018-08-08
Translations documentation

### LINKED ISSUE #696: Document how we are doing translations
What is i18n, why are we using it, how do we add a translation, what is the workflow (not google translate 😄 ).

maxneuvians:
Closes #696


sastels (approved):
👾

PR #833, opened 2018-08-08
Removed most links

### LINKED ISSUE #776: Meta Loading State
Sometimes the buttons stay in a down state without moving to the next page.

To replicate:

1. Find a place with spotty wifi
2. Load app
3. Click on "Find relevant benefits." 
ERROR <- button may stay depressed state for a long time.

### LINKED ISSUE #792: Links not working bug
Some links in the app when clicked show a hover state, but do not proceed to link. 

~~Further, some states are lost when clicked. For example, when navigating from "saved benefits" back to the directory, I clicked "Print Page" and it did not open the link, showed a hover state on the link, and removed my content filter - revealing the 21 benefits again.~~ moved to #818 

FYI this likely has something to do with moving from hrefs to next links (seems to affect links where that change has been made).

### LINKED ISSUE #837: `Link`s don't load new page if you click while current page is loading
a new page doesn't load when you click a <Link> if the current page is still loading, whereas when you click an <a> it will start loading the new page even if the current page is still loading.

smcmurtry:
Closes #837 
Closes #792 
Closes #776


maxneuvians (approved):
🦑 LGTM

PR #845, opened 2018-08-08
Fix for 839

### LINKED ISSUE #839: Improve styling of sort dropdown
New design:
screen shot 2018-08-08 at 1 11 45 pm

Current app:
screen shot 2018-08-08 at 1 10 48 pm

maxneuvians:
close #839, sorry @sastels 😢

### DIFF
--- a/components/BB.js
+++ b/components/BB.js
@@ -259,7 +259,10 @@ export class BB extends Component {
               <Grid item lg={8} md={8} sm={7} xs={12}>
                 <Grid container spacing={16}>
                   <Grid item xs={12} md={6}>
-                    <InputLabel className={classes.sortByLabel}>
+                    <InputLabel
+                      for="sortBySelector"
+                      className={classes.sortByLabel}
+                    >
                       {t("B3.Sort By")}
                     </InputLabel>
                     &nbsp;&nbsp;
--- a/components/BB.js
+++ b/components/BB.js
@@ -40,10 +40,15 @@ const styles = () => ({
     width: "100%"
   },
   formControl: {
-    minWidth: 120
+    minWidth: 120,
+    paddingTop: "4px"
   },
   sortByBox: {
-    backgroundColor: "white"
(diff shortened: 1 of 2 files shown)


components/BB.js
```diff
@@ -251,15 +259,19 @@ export class BB extends Component {
               <Grid item lg={8} md={8} sm={7} xs={12}>
                 <Grid container spacing={16}>
                   <Grid item xs={12} md={6}>
+                    <InputLabel className={classes.sortByLabel}>
```
sastels: will screen readers associate this label with the Select? I was thinking of pulling the label out like this, but thought that was a no-no.

maxneuvians: Probably not

sastels:
You may as well get rid of `classes` in `ProfileSelector` PropTypes while you're there.

image

components/BB.js
sastels: eh, it passes Axe tests.

maxneuvians: Adding a `for` to the label

sastels:
Can you center the "Sort By" vertically, and blue things up a bit?
image
vs
image

components/BB.js
```diff
@@ -259,7 +259,10 @@ export class BB extends Component {
               <Grid item lg={8} md={8} sm={7} xs={12}>
                 <Grid container spacing={16}>
                   <Grid item xs={12} md={6}>
-                    <InputLabel className={classes.sortByLabel}>
+                    <InputLabel
+                      for="sortBySelector"
```
sastels: 👍

maxneuvians:
Looks great in Chrome, don't know about  blue links, would be the only place in the app that this is blue

sastels:
hmm, yeah, looks fine in Chrome and Safari. Just funny on Firefox.

sastels (approved):
ok, close enough i guess.

PR #849, opened 2018-08-08
tweak print from favourites

### LINKED ISSUE #846: Hide profile/needs info if printing from favourites page
Also change language to "Saved benefits" from "benefits to consider".
Design:
screen shot 2018-08-08 at 2 57 25 pm

sastels:
resolves #846


smcmurtry:
It's not actually hiding the box where the profile/needs info is. I think you can just check for your `fromFavourites` flag in render and hide that whole box, rather than modifying  `profile_text` and `needs_text`.

maxneuvians (approved):
Solid! 🦑

Issue #865, opened 2018-08-09
Fr disappears on narrow phones


sastels:
being pushed off the screen by the VAC logo.

sastels:
Is there a way to crop the right side of the VAC svg?
<img width="178" alt="image" src="https://user-images.githubusercontent.com/8228248/43909843-4a45bda4-9bc9-11e8-91a4-c225debc3b9b.png">

PR #869, opened 2018-08-09
Fix for #865

maxneuvians:
smaller federal banner on phones


sastels (approved):
that banner's pretty small now, but there's crappier things about the app on a phone that narrow. not broken anymore, at least.

Issue #883, opened 2018-08-10
remove Google Maps key from code


sastels:
Only gives public access to stuff, but still shouldn't have it in the code.

https://github.com/cds-snc/vac-benefits-directory/blob/master/__tests__/components/area_office_map_test.js#L22
https://github.com/cds-snc/vac-benefits-directory/blob/master/__tests__/components/place_search_test.js#L30

sastels:
note that we also have the AirTable read key in the code #765

PR #886, opened 2018-08-10
remove airtable and google maps keys from code

### LINKED ISSUE #765: move airtable read key to env var
it's a read only key for public data, so there's perhaps no real harm in having it in github, but it's bad form, and also harder to change.

sastels:
resolves #765 and #883 

In particular, everyone will need a local copy of `AIRTABLE_READ_KEY`. I'll ping everyone when I merge this.

Note that `GOOGLE_MAPS_KEY` was already in CircleCI / `Dockerfile`


utils/airtable_es2015.js
```diff
@@ -3,7 +3,7 @@ require("isomorphic-fetch");
 exports.hydrateFromAirtable = exports.writeFeedback = undefined;
 
 var airtableConstants = require("./airtable_constants");
-var readKey = "keySzaXvONeLwsBm4"; // Read access only API key
+var readKey = process.env.AIRTABLE_READ_KEY; "keySzaXvONeLwsBm4";
```
smcmurtry: Whoops 😄

sastels: 😊

maxneuvians (approved):
🦑 LGTM

PR #889, opened 2018-08-13
Changed back behaviour, fixed tests

### LINKED ISSUE #872: after searching for a benefit from homepage, you have to hit back twice before going back to the homepage
observed on a google pixel

1. on the app homepage, enter "disability" into the `Search the benefit list` box 
2. hit enter on your phone
3. observe you're taken to the benefits directory page and "disability" is highlighted (it's entered in the search box at the top of the page)
4.  hit the back button on your phone 
5. observe that the homepage is flashed for a moment, then you are taken to the benefits directory page with no filter ("disability" is not in the search box anymore) 
6. hit the back button on your phone again 
7. observe you are taken back to the homepage 

expected that you'd be taken back to the homepage in step 5, not land on the entire benefits directory.

### LINKED ISSUE #877: URL changes for each search input character, causes UX issues
Searching on the benefits directory page happens immediately after each character, and the URL changes for each character, making navigating with the browser back and forward buttons quite difficult. 

1. https://cdsvacpoc.azurewebsites.net/benefits-directory
2. search for 'disability', observe the URL changes with each character 
3. click the back button on your browser, observe it goes back to `disabilit`, click again, goes to `disabili`, etc etc 

Expected that it would go back to the straight benefits directory. Maybe there needs to be a search button or some other way to control how the URL works.

smcmurtry:
Closes #877 
Closes #872 

Hitting the browser back button now always goes to the previous page, rather than the previous state on the same page.

One drawback is that hitting browser back from (for example) the 3rd page of guided experience will take you back to the landing page. I tried to change this to go to the previous section but it didn't work.

### DIFF
--- a/__tests__/pages/A_test.js
+++ b/__tests__/pages/A_test.js
@@ -19,10 +19,7 @@ expect.extend(toHaveNoViolations);
 jest.mock("react-ga");
 
 describe("A", () => {
-  Router.router = {
-    push: jest.fn()
-  };
-  Router.push = jest.fn();
+  Router.replace = jest.fn();
 
   let props;
   let _mountedA;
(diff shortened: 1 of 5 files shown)


sastels:
go to index page:
1) type in search
2) hit return
3) goes to BD page as expected. clicking back just clears the search, and can't get back to index.

if you hit the Search button instead of hitting return all is good.

sastels (approved):
this PR makes me happy!

PR #896, opened 2018-08-14
Selectors

### LINKED ISSUE #487: Create selectors for print url and favourites url
https://github.com/cds-snc/vac-benefits-directory/blob/ce5265eb5960707f380ebf74731dc5490950b162/components/BB.js#L141-L180
should be state based selectors

maxneuvians:
Closes #487

### DIFF
--- a/__tests__/selectors/urls_test.js
+++ b/__tests__/selectors/urls_test.js
@@ -1,4 +1,4 @@
-import { getFavouritesUrl } from "../../selectors/urls";
+import { getFavouritesUrl, getPrintUrl } from "../../selectors/urls";
 
 describe("getFavouritesUrl", () => {
   let props;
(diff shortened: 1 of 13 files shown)


maxneuvians:
Paths to test:
[X] Index page -> Favourites button (no favourites) -> Print
[X] Index page -> Favourites button (2 favourites) -> Print
[X] Index page -> Favourites button (2 favourites) -> Back

[X] Benefits directory -> Filter eligibility -> Print
[X] Benefits directory -> Filter needs -> Print
[X] Benefits directory -> Filter search -> Print
[X] Benefits directory -> Filter order -> Print
[X] Benefits directory -> Filter eligibility, Filter needs -> Print

[X] Benefits directory -> Favourites button (no favourites) -> Print
[X] Benefits directory -> Favourites button (2 favourites) -> Print
[X] Benefits directory -> Favourites button (2 favourites) -> Back

sastels:
Would be nice to have e2e testing for ☝️

maxneuvians:
Good idea

selectors/urls.js
```diff
@@ -0,0 +1,138 @@
...
+    }
+    if (closestAreaOffice.id !== undefined) {
+      url += "&closestAOID=" + closestAreaOffice.id;
+    }
+    if (selectedAreaOffice.id !== undefined) {
+      url += "&selectedAOID=" + selectedAreaOffice.id;
+    }
+    if (fromFavourites !== undefined) {
+      url += "&fromFavourites=" + fromFavourites;
+    }
```
smcmurtry, line 135: When modifying the url prop directly I believe we can do: `url.query['key'] = value`
or replace all `url.query = {key: value, ...}`. It would be nice if we could do something similar here to avoid some of the repetition.

maxneuvians: the problem is the passed arguments are not available as an iterable object, I looked into that before creating the above mess

smcmurtry (approved):
Looks good to me! Great that you added so many tests. Don't feel you need to address my other comment before merging.

selectors/urls.js
maxneuvians: I guess you can https://www.electrictoolbox.com/loop-parameters-passed-javascript-function/ but only by index not by name

smcmurtry: I was thinking more along the lines of building a `query` object with everything we want in the url, and then making another function that takes `query` as an argument and returns the url.

maxneuvians: That probably make sense - now that it is in a single place, it should be a quick refactor, I will create an issue

maxneuvians:
@sastels I tried to create test cafe e2es for this - the problem is once you hit the print page you need manual intervention to close the print dialog, otherwise it times out. Tried this solution, https://testcafe-discuss.devexpress.com/t/how-to-close-print-dialog/1107/2 no luck

__tests__/selectors/urls_test.js
```diff
@@ -0,0 +1,267 @@
...
+    );
+  });
+
+  it("adds fromFavourites string to the URL", () => {
+    params["fromFavourites"] = true;
+    expect(getPrintUrl(state, props, params)).toEqual(
+      "/print?lng=en&sortBy=relevance&fromFavourites=true"
+    );
+  });
+});
```
sastels, line 267: testing FTW! ❤️

PR #913, opened 2018-08-14
Now map zooms when user searches

### LINKED ISSUE #867: Searching on map should cause map zoom/pan to that location
This happens on mobile - android, chrome (desktop is fine, presumably because of the `search` button present). 

1. go to this page: https://cdsvacpoc.azurewebsites.net/map?lng=en
2. enter 'Ottawa' in search box
3. there is no `search` button on mobile - hit enter on your phone 
4. observe: the map stays the same 
    expected: map would zoom in to Ottawa

the list below filters to Ottawa, but users can't see immediately see it on the mobile view 
image

smcmurtry:
Closes #867


maxneuvians (approved):
Nice fix 🦑

Issue #914, opened 2018-08-14
Need to hit enter twice on place search bar before search occurs 


smcmurtry:
Clicking on a place suggestion should execute search.
Using arrow keys to select place suggestion and then hitting enter should execute search.
Typing a search term, then hitting enter should execute search.

maxneuvians:
Mistake on my part

PR #918, opened 2018-08-15
Fix for #914

maxneuvians:
Should now pan map on click search click


smcmurtry (approved):
Works very nicely!

smcmurtry:
Closes #914

PR #922, opened 2018-08-16
Fix for #750

### LINKED ISSUE #750: Map page crashes on iphone 5s
(and maybe others). Might be the geolocation code - might need to check if related objects exist before using.

maxneuvians:
Closes #750 
Changes the `values` to a similar check based on `keys` which is valid JS on iphone 5s


sastels (approved):
seems legit

PR #925, opened 2018-08-16
Streamlined polyfilling

maxneuvians:
With babel7 things were changed around `useBuiltins` (https://medium.freecodecamp.org/were-nearing-the-7-0-babel-release-here-s-all-the-cool-stuff-we-ve-been-doing-8c1ade684039). It is no longer discovering that `Object.assign` needs to be loaded with NextJS. NextJS has published solution that our solution is based on: https://github.com/zeit/next.js/tree/canary/examples/with-polyfills

### DIFF
--- a/utils/polyfills.js
+++ b/utils/polyfills.js
@@ -3,6 +3,6 @@
 // Add polyfills here. Check all available here: https://github.com/zloirock/core-js
 // This files runs at the very beginning (even before React and Next.js core)
 
-// FOLLOWING IS THE IMPORT I NEEDED FOR IE 11, 10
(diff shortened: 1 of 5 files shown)


utils/polyfills.js
```diff
@@ -1,46 +1,8 @@
...
-          }
-        }
-      }
-      return to;
-    },
-    writable: true,
-    configurable: true
-  });
-}
+// FOLLOWING IS THE IMPORT I NEEDED FOR IE 11, 10
```
sastels: maybe drop the "I"

sastels (approved):
🍻

PR #945, opened 2018-08-17
Added URL validation functionality for data validation

### LINKED ISSUE #404: Data Validation: check that VAC links are valid
check that the French and English benefit links are valid. Unfortunately this has to be done by the server (ie through `server.js`) rather than the broswer, and then the validity info (perhaps a list of `{benefitId, FailingLink}`) can be passed to the Data Validation page for display.

See the comments at the bottom of #184 for some suggestions.

maxneuvians:
Fixes #404 

There are some basic premises around this functionality:

- It should check if the URLs exist for benefits in the application
- It should not allow arbitrary URLs to be checked
- It should rate-limit benefit URL checking
- It should happen upon a user action as to not happen upon every server boot

To this end I have come up with the following flow:

- User visit the data validation page
- User clicks on "Validate" button
- Page sends each benefit ID in sequence to server
- Server looks up the URLs associated with the ID
- If the server has done this look up in the past **three** minutes
- If the server has not preformed the lookup recently then check if the URL returns a 200 status code (if it is a 300 level redirect it is technically broken, as the real URL needs to be linked for SEO)
- Server returns results is a benefits english and french URLs exist. 
- Data validation page adds links that are broken to data validation table


smcmurtry:
@maxneuvians I know you haven't requested review yet so maybe I'm jumping the gun, but I noticed the following:

- I changed one of the veterans.ca links by adding 'xxx' to the end
- I refreshed the cache and clicked the "Validate" button
- The test still passed

So I tried following the [bad link](http://www.veterans.gc.ca/eng/services/health/mental-health/understanding-mental-health/clinicsxxx) in my browser , and instead of giving me a 404 page, I was just directed to http://www.veterans.gc.ca/.

maxneuvians:
Currently you can wait three minutes, but I am also hooking the up the refresh function to clear that cache

maxneuvians:
One small change incoming

maxneuvians:
@smcmurtry looks like redirects are treated as 200 in JS fetch libraries https://fetch.spec.whatwg.org/#atomic-http-redirect-handling

maxneuvians:
Testing for this is as follows:
 - Load data validation
- Click Validate
- Ensure everything passes
- Make change in airtable to break a URL
- Click refresh in data-validation
- Refresh data-validation till new data show up
- Click on Validate
- Should show broken link

screen shot 2018-08-17 at 3 26 25 pm

__tests__/utils/url_check.js
```diff
@@ -0,0 +1,56 @@
+import { checkURL } from "../../utils/url_check";
```
sastels: should the filename be `url_check_tests.js`?

pages/data-validation.js
```diff
@@ -95,6 +145,7 @@ export class DataValidation extends Component {
       this.createData(
         "Size of Benefits Table",
         benefits.length,
+
```
sastels: 🔥 this line

utils/url_check.js
```diff
@@ -0,0 +1,42 @@
...
+      };
+
+      return urlCache;
+    }
+  }
+});
+
+var fetchUrl = async function fetchURL(url) {
+  return await fetch(url)
+    .then(res => res)
```
sastels: Is this line needed?

smcmurtry:
Okay I entered 2 broken links into airtable in the following order:

http://www.veterans.gc.ca/eng/services/health/mental-health/understanding-mental-health/clinicsxxx
abc

Only abc shows up as a broken link:
screen shot 2018-08-17 at 4 03 16 pm

maxneuvians:
@smcmurtry http://www.veterans.gc.ca/eng/services/health/mental-health/understanding-mental-health/clinicsxxx is a valid link because it redirects to a real page

smcmurtry:
@maxneuvians it redirects to a real page, but if we consider this a valid link then we won't be catching 2 of the most likely errors, which are:

- a typo in the url after the gc.ca/
- a page is taken down and they redirect to the home page

smcmurtry:
One way would be if we considered it invalid if there is any redirection. But maybe that could be a 2nd PR.

maxneuvians:
@smcmurtry I had looked into your suggestion when you mentioned the first redirect - it looks like fetch makes those requests opaque (https://fetch.spec.whatwg.org/#atomic-http-redirect-handling) so when you hit a URL do a redirect, it still returns a 200 status code instead of the expected 303 redirect code

sastels:
Do we get the url of where we were redirected to? Ie can we check to see if we got the requested page or some other page?

maxneuvians:
@sastels it's a good idea - I looked up the response object and it does not indicate that it will, but, I can try and see if it has something that says, hey I am not what you wanted.

maxneuvians:
Really good ideas @sastels and @smcmurtry. Checking the returned URL vs. the original URL found the test case and an actual error 😀 
screen shot 2018-08-20 at 3 28 21 pm

maxneuvians:
And it was actually an honest to god bug

screen shot 2018-08-20 at 3 41 41 pm

smcmurtry (approved):
Great work @maxneuvians!! It works for me when tested locally but not in the review app because refresh cache is broken.

PR #947, opened 2018-08-17
Improve airtable fetching

### LINKED ISSUE #923: keep old data if refresh fails
instead of just emptying out redux. pinging us on vac-devs would be really good too.

sastels:
fixes #923 

If the server can't fetch a table, just keep using the old data. Previously we'd throw away the old data in this situation.

to test: go rename one of the tables. after refreshing Azure that table is empty in the app, in this PR the old data is still there.

### DIFF
--- a/pages/data-validation.js
+++ b/pages/data-validation.js
@@ -221,7 +221,7 @@ DataValidation.propTypes = {
   examples: PropTypes.array.isRequired,
   i18n: PropTypes.object.isRequired,
   t: PropTypes.func.isRequired,
-  timestamp: PropTypes.object,
+  timestamp: PropTypes.number.isRequired,
   classes: PropTypes.object.isRequired,
   translations: PropTypes.array.isRequired,
   areaOffices: PropTypes.array.isRequired
--- a/server.js
+++ b/server.js
@@ -22,14 +22,20 @@ const airTable = require("./utils/airtable_es2015");
 
 const { getGithubData } = require("./utils/statistics");
 
-getAllData = async function() {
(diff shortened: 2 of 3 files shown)


pages/data-validation.js
```diff
@@ -221,7 +221,7 @@ DataValidation.propTypes = {
   examples: PropTypes.array.isRequired,
   i18n: PropTypes.object.isRequired,
   t: PropTypes.func.isRequired,
-  timestamp: PropTypes.object,
+  timestamp: PropTypes.number.isRequired,
```
sastels, line 224: noticed this was throwing a warning

server.js
```diff
@@ -22,14 +22,20 @@ const airTable = require("./utils/airtable_es2015");
 
 const { getGithubData } = require("./utils/statistics");
 
-getAllData = async function() {
+const getAllData = async function() {
   const githubData = await getGithubData();
   const airtableData = await airTable.hydrateFromAirtable();
-
   return { githubData: githubData, airtableData: airtableData };
 };
```
sastels, line 29: Just some cleanup of this function

store.js
```diff
@@ -69,7 +69,6 @@ export const reducer = (state = initialState, action) => {
 
     case "LOAD_DATA":
       newState = {
-        storeHydrated: action.data.storeHydrated || state.storeHydrated,
```
sastels, line 72: We're not using `storeHydrated` in the app 😞

maxneuvians (approved):
Looks like solid changes!

PR #953, opened 2018-08-18
General testing improvement pass

### LINKED ISSUE #628: Add tests around GE options and next section
There's a lot of logic in the page `A.js` around setting the options and the next section depending on previous selections. That logic is currently not being tested. We should take the login out of `render()` and put it in functions that we can easily test (and then write those tests).

### LINKED ISSUE #932: improve testing of map.js page and area_office_map.js component
It's currently dreadful.

maxneuvians:
Fixes #932 
Fixes #628

### DIFF
--- a/components/place_search.js
+++ b/components/place_search.js
@@ -113,8 +113,7 @@ export class PlaceSearch extends Component {
               type="text"
               placeholder={this.props.t("map.search-location-prompt")}
               className={this.props.classes.searchInputField}
-              onKeyDown={this.onKeyPress}
-              onKeyUp={this.onKeyDown}
+              onKeyDown={() => this.onKeyPress()}
             />
           </div>
           <div>
--- a/components/profile_needs_selector_mobile.js
+++ b/components/profile_needs_selector_mobile.js
@@ -132,7 +132,7 @@ export class ProfileNeedsSelectorMobile extends Component {
   }
 }
 
-const mapDispatchToProps1 = dispatch => {
+const mapDispatchToProps = dispatch => {
   return {
     setPatronType: patronType => {
       dispatch({ type: "SET_PATRON_TYPE", data: patronType });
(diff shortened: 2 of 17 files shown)


maxneuvians:
Would recommend just focusing on the small file changes to actual application code and assume tests that are passing actually added some value 😬

components/place_search.js
```diff
@@ -113,8 +113,7 @@ export class PlaceSearch extends Component {
               type="text"
               placeholder={this.props.t("map.search-location-prompt")}
               className={this.props.classes.searchInputField}
-              onKeyDown={this.onKeyPress}
-              onKeyUp={this.onKeyDown}
+              onKeyDown={() => this.onKeyPress()}
```
sastels, line 116: Why not just give it this.onKeyPress?

components/profile_needs_selector_mobile.js
```diff
@@ -132,7 +132,7 @@ export class ProfileNeedsSelectorMobile extends Component {
   }
 }
 
-const mapDispatchToProps1 = dispatch => {
```
sastels, line 135: Where did the “1” come from??

components/place_search.js
maxneuvians: `onKeyUp` wasn't doing anything, I just removed it

jest.setup.js
```diff
@@ -1,4 +1,11 @@
...
 
-configure({ adapter: new Adapter() })
+configure({ adapter: new Adapter() });
+
+const position = { coords: { latitude: 10, longitude: 10 } };
+const mockGeolocation = {
+  getCurrentPosition: async success => success(position)
+};
+
+global.navigator.geolocation = mockGeolocation;
```
sastels, line 11: Cool!

components/place_search.js
sastels: I meant why do the ()=>

sastels: Hmm I think I did that original stuff? Had to do with whether the last char you typed was used if you hit the button instead of pressing Enter?

maxneuvians: I have found that it executed all code on render that is defined without the lambda

sastels (approved):
Awesome work! 💯
🦑 🦑 🦑

maxneuvians:
@sastels Appreciate you poking at it during vacation

PR #969, opened 2018-08-21
Fixed refresh button

### LINKED ISSUE #967: Refresh cache button is broken
Does not do anything

maxneuvians:
Fixes #967 

What was happening was the data was being returned to the functionality that checks if a table is empty - without the table data, because it was still just a promise. Now it awaits till all promises are complete before it sends the data for validation.


apmcguigan (approved):
Awesome!

PR #973, opened 2018-08-21
Needs selectors 

### LINKED ISSUE #805: change needs selector to checkboxes
GE and BD

maxneuvians:
Fixes #805


smcmurtry:
On narrow screens it looks like this:
screen shot 2018-08-22 at 11 59 50 am

What about using `<Grid item>`s in the GE so it can go to a single column on narrow screens?

smcmurtry (approved):
🦑

PR #976, opened 2018-08-22
Default user location if no geolocation is called

### LINKED ISSUE #975: Map bug
Map no longer shows distance when user location is undefined - I added the bug in my test refactor

maxneuvians:
Fixes #975


smcmurtry (approved):
Nice work! I hadn't noticed that bug.

Issue #979, opened 2018-08-22
Add htmlFor attribute to FormControlLabel


maxneuvians:
Add htmlFor attribute to FormControlLabel components so that we have<label for="id"> and <input type="radio" id="id"> in the DOM

See #380

PR #990, opened 2018-08-22
Add htmlFor

smcmurtry:
I was getting AXE tests failing because my radio button/checkbox ids were not unique on /benefits-directory. This is because we are adding them 2 times - in `ProfileNeedsSelector` and `ProfileNeedsSelectorMobile`. So I added a new `parent` prop that makes the ids unique but it is an ugly solution. I would really like to get rid of `ProfileNeedsSelectorMobile`. Thoughts?


maxneuvians:
instead of a parent giving the needs selector a suffix, just have each iteration of needs button generate a random uuid for its own state so have it mount with a random uuid in state, assign that uuid as a suffix (here is a link to some uuid functions: https://stackoverflow.com/questions/105034/create-guid-uuid-in-javascript), it might even work with `Date.now()`.

maxneuvians (approved):
Nice work!

smcmurtry:
Closes #979

PR #992, opened 2018-08-23
Documentation generator and example

### LINKED ISSUE #954: Investigate code documentation and document generation
https://github.com/marborkowski/react-doc-generator

maxneuvians:
Fixes #954 

Added https://react-styleguidist.js.org/ as an automatic documentation generator for inline code comments. Added the config file and did documentation for two components as an example. 

screen shot 2018-08-23 at 11 12 37 am


components/area_office_map.js
```diff
@@ -162,12 +177,33 @@ const mapStateToProps = reduxState => {
...
+  /**
+   * Properties of the map view (Lat, Lng, Zoom)
+   */
   mapView: PropTypes.object.isRequired,
+  /**
+   * Object that contains CSS classes
+   */
   classes: PropTypes.object.isRequired,
+  /**
+   * Universal translation function
```
smcmurtry, line 205: This is definitely cool, but it will be very time consuming to add these comments to every prop in every component. Will the documentation still generate for files that don't have these comments?

maxneuvians: yes it just looks like this 

screen shot 2018-08-23 at 11 54 34 am

So it is pretty nice out of the box

sastels (approved):
🦑

Issue #920, opened 2018-08-15
Deep dive into how we should be doing page navigation


smcmurtry:
Are `next/Link`s the thing we should be using, and we were doing it wrong? Or should we use a different library? 

The objectives are:
- to have the best performance possible, as measured by page load time
- preserve correct browser history
- preserve redux between pages
- links should behave as a user would expect - clicking a link will load a new page, even if the 1st page is still loading

If we can't find a solution that meets all these objectives, we should probably stick to conventional `<a>` tags.

smcmurtry:
We significantly improved the performance of the app by reducing our bundle size. After making this improvement, I tested using `<Link>`s instead of `<a>`s in #994. On my macbook the load is now fast enough that I don't have time to click on a navigation button while it is unresponsive. But if I use chrome devtools to simulate a low-end mobile device I can.

Does this now perform well enough to switch to `<Link>`s? We should have a broader conversation about this.

sastels:
Are we good to close this issue @smcmurtry ? Seems we hve things mostly figured out.

sastels:
@smcmurtry ☝️ ?

smcmurtry:
Yes! Sorry I'm not watching my notifications very closely.

PR #994, opened 2018-08-23
[WIP] navigation experiments

smcmurtry:
Working on #920 

I only made this a PR so I can test the performance using a remote webserver.


smcmurtry:
Closed until we can make a decision on this. See #920.

PR #997, opened 2018-08-23
CSS from area_office_table into emotion

### LINKED ISSUE #991: Change component/area_office_table styling over to emotion

apmcguigan:
closes #991


smcmurtry (approved):
Looks good! There is a bug in styling of the table but it existed before this PR. I'll work on it after you merge.

PR #1008, opened 2018-08-25
Only build next on Heroku not CI or locally

maxneuvians:
Trying to fix the fact that we build next after every package install - it only needs to be built in production

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,7 +8,7 @@
   "scripts": {
     "dev": "node server.js",
     "build": "next build",
-    "postinstall": "next build",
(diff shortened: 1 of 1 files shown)


maxneuvians:
Shaves 90 seconds off CircleCI

maxneuvians:
Heroku PR build works - we made that change initially because we would timeout when we built the app in Heroku PR, so we moved it from the Procfile to packages.json. Check out https://github.com/cds-snc/vac-benefits-directory/pull/519

smcmurtry (approved):
LGTM

PR #602, opened 2018-07-17
Package upgrade for Sprint Security

maxneuvians:

PR #1013, opened 2018-08-27
Add yarn build to Dockerfile

maxneuvians:
By removing the build process from the post install process in #1008 it no longer built the site in Docker because I removed that #602, because there it was doing it twice.


smcmurtry (approved):
¯\_(ツ)_/¯

PR #1043, opened 2018-08-28
Created a container component

### LINKED ISSUE #1041: Make a Container component
Every page uses it to set the `max-width` and `margin` and `padding`. Rather than define multiple times we could just define it once and import.

smcmurtry:
Closes #1041


sastels (approved):
Much cleaner now!

PR #1059, opened 2018-08-29
conversion to emotion

### LINKED ISSUE #1029: need_tag CSS

thomascraig:
resolves #1029


sastels:
A `package-lock.json` file snuck in here and is breaking the heroku deploy - we always use `yarn` to install modules, you must have used `npm` at some point? Could you 🔥 that `package-lock.json` file?

thomascraig:
@sastels yeah my bad I tend to use npm.  lock file has been deleted just re-running unit test.

sastels:
ah, you also have a copy of the old file `pages/A.js` in there that should be 🔥 . Also, could you remove the `classes` prop from `need_tag_test.js`? (it was just there because of the `withStyles` stuff)

smcmurtry:
Also the tags are filling up the whole card width:
screen shot 2018-08-29 at 4 31 29 pm

components/need_tag.js
```diff
@@ -1,36 +1,31 @@
...
+import { css } from "react-emotion";
 
-const styles = theme => ({
-  needsTag: {
-    marginRight: theme.spacing.unit,
-    marginBottom: theme.spacing.unit,
-    color: "black",
-    borderRadius: 1,
-    display: "inline-flex",
-    padding: "4px 8px"
```
apmcguigan, line 12: missing some ;'s on the end of some of these lines. Noticed that the class is not being applied and think that could be the problem. Also worth removing  the " around inline-flex and 4px 8px

thomascraig: good call the quotes was the culprit

apmcguigan (approved):
Looks good, perfect match to master

PR #1069, opened 2018-08-30
moved all inline CSS to emotion

### LINKED ISSUE #1066: Remove inline styles from various places
- [x] components/favourites.js
- [x] componentes/federal_banner.js
- [x] compoenents/guided_experience_profile.js
- [x] components/profile_needs_selector.js
- [x] components/search.js
- [x] pages/data-validation.js
- [x] pages/index.js

I have intentionally left out a few components that include inline styles where they are used to include variable styling set by a prop.

apmcguigan:
closes #1066


sastels (approved):
Awesome work! 🦑

PR #1071, opened 2018-08-30
Remove  pageWidth

### LINKED ISSUE #789: get rid of pageWidth
we should be able to use breakpoints instead. And probably not have any flicker! 🎉 
https://material-ui.com/layout/breakpoints/

sastels:
resolves #789 

We don't need `pageWidth` much anymore since we've replaced almost all of its usage with `@media` queries. There was one spot in the `NeedButton` where it Is still required though - clicking on a `NeedButton` jumps the browser to the top of the window on desktop but not on mobile. Using materialUI's page width solution wasn't good enough here - it is only set on page load, not when the window is resized. 😞


smcmurtry (approved):
Nice work!

PR #1072, opened 2018-08-30
Removed classes from almost everywhere

### LINKED ISSUE #1070: Remove classes from component props and tests
`NeedButton` is throwing a warning because it's not getting this required prop.

smcmurtry:
Closes #1070


sastels (approved):
I wonder what other code we have in there that isn't doing anything? 😞

PR #1074, opened 2018-08-30
fix missing key

### LINKED ISSUE #1073: fix jest warnings
getting
```
Warning: Each child in an array or iterator should have a unique "key" prop
```
twice in the jest tests.

sastels:
fix #1073 

We were missing the `key` prop to `this.tableRow`, so those rows were getting an undefined key. I changed the way the key is computed in `this.tableRow`.

Line 235 in the new file is where the missing key was.

### DIFF
--- a/components/area_office_table.js
+++ b/components/area_office_table.js
@@ -139,12 +139,12 @@ export class AreaOfficeTable extends Component {
     );
   };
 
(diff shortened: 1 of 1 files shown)


smcmurtry (approved):
Nice!

PR #1091, opened 2018-09-03
Benefit additional info header

### LINKED ISSUE #929: Benefit Card headers should be able to display warnings/notifications (not just 'parent' benefits)
say for Disability Award. put that text in the (!) area. Check with Design for content

thomascraig:
resolves #929

### DIFF
--- a/components/benefit_card_additional_info.js
+++ b/components/benefit_card_additional_info.js
@@ -1,0 +1,72 @@
+import React, { Component } from "react";
+import PropTypes from "prop-types";
+import ErrorOutlineIcon from "@material-ui/icons/ErrorOutline";
+import Paper from "@material-ui/core/Paper";
+import { logEvent } from "../utils/analytics";
+import { connect } from "react-redux";
+import { css } from "react-emotion";
+
+const cardTop = css`
+  background-color: #f1f7fc;
+  border-radius: 0px;
+  border-bottom: 1px solid #8b8b8b;
+  padding: 15px 15px 15px 10px;
+  display: flex;
+  justify-content: center;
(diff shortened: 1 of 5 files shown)


components/benefit_cards.js
```diff
@@ -11,11 +11,12 @@ import NeedTag from "./need_tag";
 import { css } from "react-emotion";
 import CardFooter from "./card_footer";
 import BenefitCardHeader from "./benefit_card_header";
+import BenefitCardHeaderMoreInfo from "./benefit_card_additional_info";
 
 const button = css`
   background-color: #3e57e2 !important;
   color: white !important;
-  text-align: right !important;
+  benefit_card_additional_infotext-align: right !important;
```
sastels: I think this is a paste where it shouldn't be?

components/benefit_card_additional_info.js
```diff
@@ -0,0 +1,72 @@
...
+  transform: scale(0.9);
+  color: #434343;
+`;
+const headerDesc = css`
+  flex-grow: 1;
+  color: #434343;
+`;
+
+export class BenefitCardHeaderMoreInfo extends Component {
+  logExit = url => {
```
sastels: you don't need the `logExit` function here, it's just used to log when the user clicks on a VAC site url in the other header.

sastels (approved):
Minor comments - awesome work! 🦑

components/benefit_card_additional_info.js
thomascraig: @sastels just seeing these comments now, were the files updated?

PR #1093, opened 2018-09-04
Fix filter reset of NeedButton

### LINKED ISSUE #1092: Reset button toggle broken
Reset doesn't toggle checks in filter by category :bug:

To reproduce click any options in Filter by Category. 
Then click reset.

sastels:
fixes #1092 

We were treating `undefined` as `false`, but MaterialUI's `CheckBox` was not interpreting it this way.


smcmurtry (approved):
Nice fix!

PR #1138, opened 2018-09-06
resized breadcrumb

thomascraig:
resolves #1138 updated h1 to use h1 component and resized typography.

### DIFF
--- a/components/guided_experience.js
+++ b/components/guided_experience.js
@@ -9,6 +9,7 @@ import { css } from "react-emotion";
 import Container from "../components/container";
 import Header1 from "../components/header1";
 import Header2 from "../components/header2";
+import { globalTheme } from "../theme";
 
 const root = css`
   border: solid 1px grey;
@@ -41,13 +42,14 @@ const nextButton = css`
 const title = css`
   font-size: 14px !important;
   line-height: 21px;
-  width: auto;
+  width: 100px;
+  display: inline-block;
 `;
 
 const jumpButton = css`
   font-size: 14px !important;
   line-height: 21px;
-  color: #006cc9;
+  color: ${globalTheme.colour.cerulean};
 `;
 
 const comma = css`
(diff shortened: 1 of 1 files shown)


components/guided_experience.js
```diff
@@ -37,12 +39,15 @@ const nextButton = css`
...
+  font-size: 14px !important;
+  line-height: 21px;
+  width: auto;
 `;
 
 const jumpButton = css`
-  font-size: 1.5em;
+  font-size: 14px !important;
+  line-height: 21px;
+  color: #006cc9;
```
sastels: you can use `theme.colour.cerulean` for this colour

sastels:
Can you get the breadcrumbs to be closer to "My selection"? From https://app.zeplin.io/project/5acbd659dc553ec84805afd4/screen/5b758b9388a3e676eae08241

image

components/guided_experience.js
thomascraig: ok, so all the colors in zeplin are added to the globalTheme?

sastels (approved):
minor comments, 🦑

components/guided_experience.js
sastels: yeah, I added that yesterday I think?

thomascraig: @sastels should I replace all the color ref's from hex vs globalTheme within the file?
eg: background-color: #39824d !important;   used on the next button

sastels: yeah, any ones you find in `theme.js`. `#39824d` is `fernGreen` :)

thomascraig:
@sastels I decreased the grid size from 3 to 2, will try size 1 but thinking it will be too small.

Attached is with grid size from 2 to 1, get some wrapping happening due to size restraints.
screencapture-localhost-3000-guided-2018-09-06-16_26_25

sastels:
ah fair enough. Maybe just leave it as it was before, we can always open a new issue to look at it later.

thomascraig:
@sastels I can tweak it more to look like the design, will just need to utilize more css.

sastels:
up to you 😄

PR #1144, opened 2018-09-06
Tweak header anchor link

### LINKED ISSUE #1136: Change hover
Remove that background hover on Save for Later, Reset and the Back buttons, instead only the text and icon should change colour

sastels:
Tweaked the HeadAnchorLink so that `onClick` works, and use for reset button in benefits directory.

resolves #1136 (last remaining bit was the reset button in benefits directory)


smcmurtry (approved):
Nice work!

PR #1149, opened 2018-09-07
restyle back button

### LINKED ISSUE #1132: Change back button in guided experience into link
Once # 1125 merges, change the GE back button into a HeaderAnchorLink.

sastels:
resolves #1132 

Had issues using the `HeaderAnchorLink` component so just restyled in place to look the same


smcmurtry (approved):
Looks good! As discussed IRL let's try to modify the `HeaderAnchorLink` component so it can be used here.

Issue #1155, opened 2018-09-07
Map search: clicking search button does not execute a search


smcmurtry:
Discovered this while swapping out the search box component. Tried to fix but couldn't.

sastels:
This is quite a rabbithole. Was it working before you swapped out / restyled things?

PR #1157, opened 2018-09-07
Swap place searchbar

### LINKED ISSUE #1156: Swap out map search box for the new component
See designs: https://app.zeplin.io/project/5acbd659dc553ec84805afd4/screen/5b5899c9a1987c122c011240

smcmurtry:
Closes #1156
Discovered an existing bug that should be fixed in a future PR: #1155

### DIFF
--- a/components/search_box.js
+++ b/components/search_box.js
@@ -30,7 +30,7 @@ const InputSearchBox = styled("input")({
   WebkitAppearance: "none",
   ":focus": {
     marginRight: "3px",
-    outline: `3px solid ` + globalTheme.colour.gdsYellow,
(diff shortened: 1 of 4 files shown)


components/search_box.js
```diff
@@ -51,7 +51,6 @@ const SearchButton = styled("button")({
   backgroundRepeat: "no-repeat",
   backgroundPosition: "2px 50%",
   ":focus": {
-    outline: `3px solid ` + globalTheme.colour.gdsYellow,
```
sastels, line 54: We might want to leave something like this in re the browser difference?

sastels (approved):
LGTM, just that browser focus border difference

PR #1158, opened 2018-09-07
switch HeaderAnchorLink to button and use for GE Back

### LINKED ISSUE #1154: make HeaderAnchorLink into a button instead of a link
and use it for the back button on the guided experience page.

sastels:
resolves #1154

### DIFF
--- a/components/guided_experience.js
+++ b/components/guided_experience.js
@@ -1,7 +1,6 @@
 import React, { Component } from "react";
 import PropTypes from "prop-types";
 import { Grid } from "@material-ui/core";
-import OldButton from "@material-ui/core/Button";
 import ArrowBack from "@material-ui/icons/ArrowBack";
 import { connect } from "react-redux";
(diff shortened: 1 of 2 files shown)


components/header_anchor_link.js
```diff
@@ -52,17 +54,30 @@ class HeaderAnchorLink extends Component {
       target,
       size
     } = this.props;
+
+    let buttonOnClick;
+    if (href) {
```
smcmurtry, line 59: awesome!

components/guided_experience.js
```diff
@@ -88,7 +77,7 @@ export class GuidedExperience extends Component {
         >
           <ArrowBack />
           {t("back")}
-        </OldButton>
+        </HeaderAnchorLink>
```
smcmurtry, line 80: Can we change the name now that it's not an anchor? 😄

smcmurtry (approved):
nice work

PR #1190, opened 2018-09-26
Moved showQuestion function to utils/common

### LINKED ISSUE #1189: Move the show/hide question logic to utils/ so we can use it in the guided experience

smcmurtry:
Closes #1189


sastels (approved):
🐝

PR #1195, opened 2018-09-27
Generalize question save to redux

sastels:
use a general "save question response" approach rather than hard coded "set patron type", "set service type", etc.


smcmurtry (approved):
Looks great! I might have called it `saveToRedux` so we could use it for other things but we can change the name later if we start doing that.

sastels:
👍 I figured I’d make it explicit that (at least right now) we’re saving the question responses. Codes complicated enough 😬

Issue #1107, opened 2018-09-04
Tweak Benefit Card headers and footers


sastels:
Benefit card top and bottom attachments should look like part of the card themselves. Currently you can see a part of the rounded corners underneath the blue. 

Also bottom should no longer be blue.

smcmurtry:
Design: https://app.zeplin.io/project/5acbd659dc553ec84805afd4/screen/5b3e525390e27e2f31bf67d4

PR #1207, opened 2018-09-28
Made the footer more like the designs

smcmurtry:
See: https://app.zeplin.io/project/5acbd659dc553ec84805afd4/screen/5b3e525390e27e2f31bf67d4

Part of #1107


sastels (approved):
LGTM. Noticed that the line separating the footer from the card disappears when expanded - is this intentional?

sastels:
huh, the line disappears in master too. 🤷‍♂️

smcmurtry:
It seems like the line gets thinner in the designs but doesn't totally disappear.

PR #1210, opened 2018-10-01
Clear hidden questions

sastels:
in ProfileSelector, clear the selected values of questions that are hidden.
for example, I've change the `QuestionClearLogic` table so that if status is `deceased`, going for family to service-person only clears the `statusAndVitals`. So in master, `serviceHealthIssue` remains filled in, even though it's not visable in the app.

In this PR, `serviceHealthIssue` gets cleared because it is hidden.

This will make it easier to add / change questions in the future.

### DIFF
--- a/__tests__/components/profile_selector_test.js
+++ b/__tests__/components/profile_selector_test.js
@@ -6,6 +6,10 @@ import questionsFixture from "../fixtures/questions";
 import multipleChoiceOptionsFixture from "../fixtures/multiple_choice_options";
 import questionDisplayLogicFixture from "../fixtures/question_display_logic";
 import questionClearLogicFixture from "../fixtures/question_clear_logic";
+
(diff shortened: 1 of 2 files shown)


__tests__/components/profile_selector_test.js
```diff
@@ -6,6 +6,10 @@ import questionsFixture from "../fixtures/questions";
 import multipleChoiceOptionsFixture from "../fixtures/multiple_choice_options";
 import questionDisplayLogicFixture from "../fixtures/question_display_logic";
 import questionClearLogicFixture from "../fixtures/question_clear_logic";
+
+jest.unmock("../../utils/common");
+const common = require.requireActual("../../utils/common");
```
sastels, line 11: Coding by Google

smcmurtry (approved):
Nice work!

Issue #1221, opened 2018-10-02
Update CircleCI to do a container scan on deploy


maxneuvians:
Use clair to scan the container before it is pushed to docker hub, remove the ecs stuff.

maxneuvians:
Closed in #1225

PR #1225, opened 2018-10-02
Added clair scanning to container

maxneuvians:
Addresses #1221. This adds a step to the deployment of the docker image that scan the container against the Clair Database of vulnerabilities. Currently the deploy will fail if there is a vulnerability of High or above.


sastels:
So does the docker image currently pass? 

If it fails will we get some info on what the problem is? (via a link to the clair scanning website or info in the text fail dump?)

maxneuvians:
Yes it passes - if it does not it will list all the vulnerabilities in the STDOUT of the CI, like lint or jest

sastels (approved):
🦑 🐝

PR #1226, opened 2018-10-02
Hover for cookies disabled

### LINKED ISSUE #1222: Tooltip hover if cookies disabled
hover for save for later button if cookies disabled

image

https://app.zeplin.io/project/5acbd659dc553ec84805afd4/screen/5b758c1e04618a550dec9617

https://trello.com/c/0bIq7XEM/21-user-tries-to-save-a-benefit-for-later-without-cookies-enabled

sastels:
If the user has cookies disabled, disable the "Save for Later" button and add a hover.

resolves #1222

### DIFF
--- a/components/favourite_button.js
+++ b/components/favourite_button.js
@@ -32,13 +32,14 @@ const bookmarkIcon = css`
   }
 `;
 const tooltiptext = css`
+  font-size: 14px;
+  font-weight: normal;
+  text-align: left;
   visibility: hidden;
   width: 160px;
-  font-size: 14px;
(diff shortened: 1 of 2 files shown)


sastels:
[User tries to save a benefit for later without cookies enabled](https://trello.com/c/0bIq7XEM/21-user-tries-to-save-a-benefit-for-later-without-cookies-enabled)

smcmurtry (approved):
Wow, great job! I like that you have a css-only solution rather than a library or something. 
Nitpicks: Could be its own component, text could should be aligned right, not bold and more padding. But all that could be a separate PR.

PR #1227, opened 2018-10-02
Swapped out our button component

smcmurtry:
I swapped the get directions button to our button component. 

I just realized that the button in the designs for the page is 35px high, while the styleguide only defines buttons that are 44px or 59px high. 

@michellengaiCDS do you think I should make a new smaller version of our button? If so could you add it to the style guide?

### DIFF
--- a/components/area_office_map.js
+++ b/components/area_office_map.js
@@ -1,8 +1,7 @@
 import React, { Component } from "react";
 import PropTypes from "prop-types";
 import { connect } from "react-redux";
-import { Button } from "@material-ui/core/";
-import { KeyboardBackspace } from "@material-ui/icons";
+import Button from "./button";
 import { css } from "react-emotion";
 
 import {
@@ -14,7 +13,7 @@ import {
 
 const officeAddress = css`
   font-size: 12px;
-  font-family: 'Merriweather', Georgia, serif;
+  font-family: "Merriweather", Georgia, serif;
 `;
 const officeInfo = css`
   font-color: black;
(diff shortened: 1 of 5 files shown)


smcmurtry:
Closes #1260 
Closes #1258

components/area_office_map.js
```diff
@@ -14,35 +13,17 @@ import {
 
 const officeAddress = css`
   font-size: 12px;
-  font-family: 'Merriweather', Georgia, serif;
+  font-family: "Merriweather", Georgia, serif;
 `;
 const officeInfo = css`
   font-color: black;
   background-color: white;
+  margin: 2px;
```
smcmurtry, line 18: this is because the focus outline of the button was getting cut off

components/area_office_map.js
```diff
@@ -14,35 +13,17 @@ import {
...
-const rightArrowIcon = css`
-  -moz-transform: scaleX(-1);
-  -o-transform: scaleX(-1);
-  -webkit-transform: scaleX(-1);
-  transform: scaleX(-1);
-  float: left;
-  filter: FlipH;
-  -ms-filter: Flip;
-  padding-right: 10px;
+  font-family: "Merriweather", Georgia, serif;
```
sastels: don't need the quotes if it's just one word (I just learned about font families today! :))

sastels (approved):
looking sharp! 👍 🐝 🦑

components/area_office_map.js
```diff
@@ -117,28 +90,21 @@ export class AreaOfficeMap extends Component {
...
-                    <div className={officeName}>
-                      {d["name_" + t("current-language-code")]}
-                    </div>
-                    <br />
-                    <div className={officeAddress}>
-                      {d["address_" + t("current-language-code")]}
-                    </div>
+                    <Header4>{d["name_" + t("current-language-code")]}</Header4>
                     <br />
+                    <Body>{d["address_" + t("current-language-code")]}</Body>
```
sastels, line 95: 👍 good catch!

PR #1228, opened 2018-10-03
Check for re-enabled cookies when hovering over favourite button

sastels:
1) disable cookies
2) load BD page. Cookies disabled alert shows, hover shows when over a benefit's favourite button
3) enable cookies
4) mouse over a favourite button. The hover should not appear, the alert should disappear, and the user should be able to click to add the benefit to favourites.


smcmurtry (approved):
Works for me in firefox! 🦑

PR #1233, opened 2018-10-03
Changed dropdown to be closer to the designs

### LINKED ISSUE #1214: Change sort by dropdown
screen shot 2018-10-02 at 10 28 00 am

smcmurtry:
Closes #1214

Still need to change the arrow to be like the designs. I will do that in a separate PR.


theme.js
```diff
@@ -32,7 +34,8 @@ const theme = {
     tornadoRed: "#ea4335",
     govukYellow: "#FFBF47",
     white: "#ffffff",
-    cardGrey: "#f3f3f3"
+    cardGrey: "#f3f3f3",
+    black: "#000000"
```
sastels, line 38: Are you using black in this PR or was it just since its a good thing to have?

__tests__/components/BB_test.js
```diff
@@ -95,7 +95,7 @@ describe("BB", () => {
   });
 
   it("has a sortBy selector", () => {
-    expect(shallow_BB().find("#sortBySelector").length).toEqual(1);
+    expect(mounted_BB().find("#sortBySelector").length).toEqual(2);
```
sastels, line 98: why did this change to 2? mount vs shallow?

sastels (approved):
minor questions, looks sharp! 🦑

__tests__/components/BB_test.js
smcmurtry: not sure - I chalked it up to the issue where enzyme sometimes returns multiple copies of the same thing.

theme.js
smcmurtry: I think I was using it at some point but no longer. But it's probably good to have in there.

PR #1236, opened 2018-10-03
Fix mobile reset button

### LINKED ISSUE #1232: Mobile Reset button only clears needs
doesn't clear profile filters.

sastels:
fixes #1232 

We hadn't refactored `profile_needs_selector_mobile.js`. When I did that the problem went away `¯\_(ツ)_/¯`


smcmurtry (approved):
Nice work! Can't believe we forgot to do that...

PR #1251, opened 2018-10-04
Combine card header components

### LINKED ISSUE #1234: Combine `BenefitCardHeader` and `BenefitCardHeaderMoreInfo`
Make this one component that we can pass different text (including `<a>`s) into. This will help make it easier to keep the styling consistent.

smcmurtry:
Closes #1234


components/benefit_card_header.js
```diff
@@ -27,84 +28,38 @@ const headerDesc = css`
...
     );
-
-    if (
+    const includeParentInfo =
       parentBenefits.length > 0 &&
-      benefit.availableIndependently === "Requires Gateway Benefit"
-    ) {
+      benefit.availableIndependently === "Requires Gateway Benefit";
+    const includeImportantInfo =
+      benefit.noteEn !== undefined && benefit.noteFr !== undefined;
```
sastels: We could use `benefit.noteEn && benefit.noteFr` here to also catch the case where one of them is "".

sastels (approved):
Nicely done! Breaking up a long render into components like this makes it much easier to see what's going on. 👍

Issue #1246, opened 2018-10-04
Refactor the parsing url code


sastels:
This is in `pages/_app.js`. Note that the code uses the `SET_PATRON_TYPE`, etc reducers that we deleted in our refactoring :( . This is almost certainly what caused #1229

sastels:
removed the reference to `SET_PATRON_TYPE`, etc in #1252 . Still need to refactor the code though.

sastels:
![](https://github.trello.services/images/mini-trello-icon.png) [User answers as little profile questions as possible](https://trello.com/c/MF0pMzGT/20-user-answers-as-little-profile-questions-as-possible)

PR #1252, opened 2018-10-04
fix Benefits Directory reload

### LINKED ISSUE #1250: Reloading the benefits directory clears the profile filters
This is because our refactor deleted the `SET_PATRON_TYPE`, etc reducers, but `_app.js` uses these reducers 😞

sastels:
fixes #1250 and partially deals with #1246 

We need to better refactor this code to not need the list of questions, but at least now it's not broken anymore.

### DIFF
--- a/pages/_app.js
+++ b/pages/_app.js
@@ -19,11 +19,31 @@ export default withRedux(initStore)(
         });
       }
 
+      // this has to be refactored
+      // unfortunately when this runs, currentReduxState.questions is empty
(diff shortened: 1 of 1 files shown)


smcmurtry (approved):
LGTM 🦑

sastels:
[User answers as little profile questions as possible](https://trello.com/c/MF0pMzGT/20-user-answers-as-little-profile-questions-as-possible)

PR #1261, opened 2018-10-05
Switched the dropdown arrow

### LINKED ISSUE #1242: Change arrow on sorting dropdown
https://app.zeplin.io/project/5acbd659dc553ec84805afd4/screen/5b45ad8f22c744f901194d96

smcmurtry:
Closes #1242

### DIFF
--- a/components/dropdown.js
+++ b/components/dropdown.js
@@ -3,6 +3,8 @@ import PropTypes from "prop-types";
 import { css } from "react-emotion";
 import { globalTheme } from "../theme";
 
+const cerulean = globalTheme.colour.cerulean.replace("#", "%23"); //The character # is reserved in URLs as the start of a fragment identifier. You must encode this as %23 for the URL to be valid.
+
 const wrapper = css`
   display: flex;
 `;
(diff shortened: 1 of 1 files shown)


components/dropdown.js
```diff
@@ -19,9 +21,9 @@ const right = css`
 const selectStyle = css`
   -webkit-appearance: none;
   -moz-appearance: none;
-  background-position: right 50%;
+  background-image: url('data:image/svg+xml;utf-8,<svg xmlns="http://www.w3.org/2000/svg" width="24" height="24" viewBox="0 0 24 24"><path d="M16.59 8.59L12 13.17 7.41 8.59 6 10l6 6 6-6z" fill="${cerulean}"/><path d="M0 0h24v24H0z" fill="none"/></svg>');
```
sastels, line 24: How does this work, are we downloading something from the web here?
Why not just use the `ExpandMoreIcon` as we do in the card footer?

smcmurtry: My fuzzy understanding is that you always include background images in css with this url function. If the image is hosted somewhere you pass in that url, if not you can just put the image source code in (which is what I've done here).

I believe the `xmlns="http://www.w3.org/2000/svg"` doesn't download anything - it just indicates what version of the svg specification we're using.

sastels: ah, I see. interesting. 👍 
Should we replace our material ui `ExpandMoreIcon` with this one?

sastels (approved):
👍 cool!

PR #1267, opened 2018-10-09
Custom paper component swap

### LINKED ISSUE #1259: Make our own paper component
It should have pointy corners and the same shadow as the MUI paper. If we use the same component everywhere and maybe just change the padding this would solve a couple problems:
-currently a lot of inconsistency in pointy vs rounded corners of paper in the app
-inconsistency in shadows, border (guided experience paper is missing the shadows)

smcmurtry:
Closes #1259 

Ended up cleaning up lots of stray margins, and putting the 2nd `<div>` we use for setting the page width into `Container`.

### DIFF
--- a/__tests__/components/paper_test.js
+++ b/__tests__/components/paper_test.js
@@ -1,0 +1,23 @@
+import React from "react";
+import { mount } from "enzyme";
+import Paper from "../../components/paper";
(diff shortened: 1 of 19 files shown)


sastels:
Did you add unit tests for `Paper`?

sastels (approved):
LGTM 🐝 🦑

PR #1282, opened 2018-10-10
Colors using globalTheme instead of hex

### LINKED ISSUE #1257: Replace all the hex coded colours in components with named colours from theme.js

apmcguigan:
closes #1257


theme.js
```diff
@@ -1,5 +1,9 @@
 const theme = {
   fontFamily: "Merriweather",
+  maxWidth: "1200px",
+  margin: "0 auto",
+  paddingLeft: "16px",
+  paddingRight: "16px",
```
sastels: I think these four lines were deleted from master in a prior PR, possibly after you started your branch.

sastels (approved):
minor comment, good work! 🦑

theme.js
smcmurtry: Yes that styling should just be in `Container` now

PR #1286, opened 2018-10-10
Created a tooltip component

### LINKED ISSUE #1285: Make a tooltip component
This will make the tooltip reusable and make it easy to do #1272

smcmurtry:
Closes #1285 

I decided this was probably better than putting the styles in theme.js.

Still to do: use it for the health issue question.


components/tooltip.js
```diff
@@ -0,0 +1,65 @@
...
+import PropTypes from "prop-types";
+import { css } from "react-emotion";
+import { globalTheme } from "../theme";
+
+const tooltipTextStyle = css`
+  font-size: 14px;
+  font-weight: normal;
+  text-align: left;
+  visibility: hidden;
+  width: 160px;
```
sastels, line 11: We might want the tooltip text width to be configurable. Fine to wait and see when we use it for the health question.

components/tooltip.js
```diff
@@ -0,0 +1,65 @@
...
+      border-style: solid;
+      border-color: ${
+        globalTheme.colour.paleGrey
+      } transparent transparent transparent;
+  }
+`;
+
+class Tooltip extends Component {
+  render() {
+    const { children, showTooltip, tooltipText } = this.props;
```
sastels: a `disabled` option might be better (more commonly used in other places)  than `showTooltip`.

sastels (approved):
A couple questions, but this is awesome! 💯

PR #1293, opened 2018-10-11
Health issue tooltip2

### LINKED ISSUE #1272: Add tooltips to Benefits Directory
When #1271 merges, add the tooltip as a hover to the Benefits Directory questions.

smcmurtry:
Closes #1272

### DIFF
--- a/components/tooltip.js
+++ b/components/tooltip.js
@@ -3,7 +3,7 @@ import PropTypes from "prop-types";
 import { css } from "react-emotion";
 import { globalTheme } from "../theme";
 
-const padding = 10;
+const padding = 16;
 
 class Tooltip extends Component {
   tooltipTextStyle = css`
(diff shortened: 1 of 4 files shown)


components/tooltip.js
```diff
@@ -3,52 +3,54 @@ import PropTypes from "prop-types";
...
-  :hover {
-    .${tooltipTextStyle} {
-      visibility: visible;
+class Tooltip extends Component {
+  tooltipTextStyle = css`
+    font-size: 14px;
+    font-weight: normal;
+    text-align: left;
+    visibility: hidden;
+    width: ${this.props.width}px;
```
sastels, line 14: I like the way you did this 👍

components/tooltip.js
```diff
@@ -57,9 +59,16 @@ class Tooltip extends Component {
...
+  disabled: PropTypes.bool,
+  width: PropTypes.number,
   tooltipText: PropTypes.string,
   className: PropTypes.string
 };
 
+Tooltip.defaultProps = {
+  disabled: false,
+  width: 150,
+  tooltipText: "tooltip text"
```
sastels: Why not make tooltipText required with no default? Seems like using the default value here would always be a bug in any use of the component.

sastels (approved):
Minor comment, looks awesome! 🦑

components/tooltip.js
smcmurtry: good point! I'll change that.

PR #1307, opened 2018-10-16
remove some unused props

sastels:
Just remove a few more props we didn't need.

### DIFF
--- a/components/card_footer.js
+++ b/components/card_footer.js
@@ -181,7 +181,6 @@ export class CardFooter extends Component {
                         benefit={cb}
                         t={this.props.t}
                         key={cb.id}
-                        showFavourite={this.props.showFavourite}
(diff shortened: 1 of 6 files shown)


components/card_footer.js
```diff
@@ -228,7 +226,6 @@ const mapStateToProps = reduxState => {
   };
 };
 CardFooter.propTypes = {
-  showFavourite: PropTypes.bool.isRequired,
```
smcmurtry, line 231: Hmm, I guess that was a relic of when we had favouriting on the embedded cards.

smcmurtry (approved):
Nice! 🦑

components/card_footer.js
sastels: 😞 yeah

PR #1308, opened 2018-10-16
make Airtable base key configurable

sastels:
allow us in the future to deploy master and use with a different Airtable base without code changes.


smcmurtry (approved):
Good idea.

PR #1321, opened 2018-10-17
make options mandatory

### LINKED ISSUE #1319: Make options in radio_selector.js` mandatory
Right now the `options` prop is allowed to be empty, in which case we use
`Array.from(new Set(this.props.eligibilityPaths.map(ep => ep[this.props.selectorType])))`
But I think we always pass the props in now? Ensure this and get rid of the above set.

sastels:
resolves #1319

### DIFF
--- a/components/radio_selector.js
+++ b/components/radio_selector.js
@@ -79,13 +79,7 @@ export class RadioSelector extends React.Component {
   render() {
     const guid = uuidv4();
 
-    const options = this.props.options
-      ? this.props.options
-      : Array.from(
-          new Set(
-            this.props.eligibilityPaths.map(ep => ep[this.props.selectorType])
-          )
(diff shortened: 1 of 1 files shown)


components/radio_selector.js
```diff
@@ -79,13 +79,7 @@ export class RadioSelector extends React.Component {
...
     const guid = uuidv4();
 
-    const options = this.props.options
-      ? this.props.options
-      : Array.from(
-          new Set(
-            this.props.eligibilityPaths.map(ep => ep[this.props.selectorType])
-          )
-        ).filter(st => st !== "na");
+    const options = this.props.options;
```
smcmurtry: you could just add it to the assignment on line 84

smcmurtry (approved):
🦑

PR #1322, opened 2018-10-17
Changed how we're handling GE breadcrumbs so no special logic is required for service health issue

### LINKED ISSUE #1311: refactor serviceHealthIssue values
So the question gets values "true" or "false". However, we often display the strings "GE.has service related health issue" or "GE.has service related health issue". This breaks the new model for how we get and use the questions from AirTable.

This happens in `pages/print.js` and `components/guided_experience.js`

smcmurtry:
Closes #1311


sastels:
Can you change it so that if `ge_breadcrumb_*` is empty / undefined it defaults to `display_text_*`?

sastels (approved):
minor suggestion, good work!

PR #1325, opened 2018-10-18
Log page render times

### LINKED ISSUE #82: Measure page render times
Track page render times as a metric

maxneuvians:
Closes #82 

Adds a cloud event to log page render times in ms.


sastels (approved):
👍 Thanks Max! 🦑

PR #1350, opened 2018-10-23
 focus is now returned to the top of the page when you click next/back in guided experience

### LINKED ISSUE #1318: Focus stays on Next button when you arrive at a new page in guided experience
The solution is to programmatically focus on the outermost `div` when the component did mount. Assign a `ref` prop like this: `<div ref={this.divRef}>`. Then set the focus like this: `this.divRef.current.focus()`.

screen shot 2018-10-17 at 1 59 22 pm

smcmurtry:
Closes #1318

### DIFF
--- a/components/guided_experience.js
+++ b/components/guided_experience.js
@@ -38,10 +38,6 @@ const body = css`
 `;
 
 export class GuidedExperience extends Component {
-  constructor(props) {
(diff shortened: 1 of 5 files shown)


sastels (approved):
neat! 🦑

PR #1360, opened 2018-10-23
Clean up fixtures

### LINKED ISSUE #1342: improve fixtures
So our `eligibilityPathsFixture` doesn't contain the field we are now using, `requirements`. Other fixtures are probably in poor shape as well. It would be very useful to go through them all and clean them up.

sastels:
resolves #1342 
clean up our fixtures. delete ones we don't use anymore, change ids and other field values to more readable values, and then fix the tests to use the new fixtures.


__tests__/selectors/benefits_test.js
```diff
@@ -18,84 +22,32 @@ describe("Benefits Selectors", () => {
     };
     state = {
       questions: questionsFixture,
-      benefits: [
```
smcmurtry, line 21: wow, I didn't know we were doing that...

smcmurtry (approved):
Nice work!!

__tests__/selectors/benefits_test.js
sastels: :(

PR #1361, opened 2018-10-23
Removed override which disabled focus styles on checkboxes

### LINKED ISSUE #1344: Checkboxes in /benefits-directory and /guided have no focus style
This prevents users from interacting using the keyboard alone.

brdunfield:
Resolves #1344

### DIFF
--- a/components/layout.js
+++ b/components/layout.js
@@ -32,7 +32,6 @@ const backgoundColour2 = css`
   background-color: ${globalTheme.colour.greyishBrown};
 `;
 const theme = createMuiTheme({
-  overrides: {},
   palette: {
     primary: {
       main: globalTheme.colour.cerulean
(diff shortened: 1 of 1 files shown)


components/layout.js
```diff
@@ -32,12 +32,7 @@ const backgoundColour2 = css`
   background-color: ${globalTheme.colour.greyishBrown};
 `;
 const theme = createMuiTheme({
-  overrides: {
-    MuiTouchRipple: {
-      childPulsate: { opacity: 0 },
-      ripplePulsate: { opacity: 0 }
-    }
-  },
+  overrides: {},
```
sastels: Would it work if we totally removed the `overrides` property?

sastels (approved):
nice! Not sure why we put those overrides in. 🤷‍♂️

PR #1365, opened 2018-10-24
Hide irrelevant questions

### LINKED ISSUE #1323: hide irrelevant questions
Based on the eligibility paths we should be able to determine which questions should be hidden based on previous input. For example, if patronType is "organization", then all subsequent questions should be hidden, as none of the options for those questions make a difference.

Currently we encode this logic in the `questionDisplayLogic` table, but this is brittle, and we should compute it directly from the `eligibilityPaths` table. This could be done by the server, ie the AirTable code running on the server could essentially create / add to the `questionDisplayLogic` table before returning data to the main `server.js` code.

sastels:
resolves #1323


__tests__/utils/common_test.js
```diff
@@ -1,47 +1,80 @@
...
+  beforeEach(() => {
+    reduxState = {
+      eligibilityPaths: eligibilityPathsFixture,
+      multipleChoiceOptions: multipleChoiceOptionsFixture
+    };
+  });
+
+  it("returns false if question is not relevant", () => {
+    const profileFilters = {
+      patronType: "p2"
```
smcmurtry, line 18: Using p1, p2 etc makes it harder to understand what the tests are doing without looking at the fixture.

smcmurtry (approved):
LGTM! 🦑

__tests__/utils/common_test.js
smcmurtry: nvm, you convinced me :)

sastels: woot!

PR #1373, opened 2018-10-25
Refactor card footer

### LINKED ISSUE #1349: refactor card_footer.js
The `getBenefitIds` function relies on the depricated `patronType` field in the eligibilityPaths, rather than the new `requirements` field.

sastels:
fixes #1349 
Use `requirements` column to determine Veteran / family benefits instead of the deprecated `patronType` column. Also remove `patronType` from the `eligibilityPaths` fixture.


smcmurtry (approved):
Good idea 🦑

PR #1379, opened 2018-10-26
HeaderButton components now use anchor tags

### LINKED ISSUE #1351: HeaderButtons that look like links do not behave as links
For example, the user cannot right click and open them in a new tab, the URL of the link is not visible in the browser before clicking, etc.

This is because these elements are styled buttons. We should probably find a way to make them anchors, most critically links that exit the app such as the MyVAC link in the favourites page.

There may be issues around Redux and maintaining data if "internal" links are changed to anchors.

brdunfield:
Perhaps the name should be refactored to "HeaderLink"?

resolves #1351 , HeaderButtons now use anchor tags and their usage (links) now match their styling.


sastels:
If I click on "Print this page" it opens the print page both in the current tab and in a new tab

brdunfield:
@sastels should be fixed now

brdunfield:
Working on making data persist through Redux - this branch is currently broken.

brdunfield:
Everything should be working correctly, but this raises #1388 , as some existing HeaderButtons are now anchors that shouldn't be. I will work on that issue on a new branch / PR once this is approved.

components/header_button.js
```diff
@@ -61,22 +45,30 @@ class HeaderButton extends Component {
...
       >
         {arrow === "back" ? <ArrowBack /> : null}
         {children}
         {arrow === "forward" ? <ArrowForward /> : null}
-      </button>
+      </a>
     );
+    // If the useLink prop is specified, wrap the anchor in a next Link to preserve data in Redux
+    if (useLink) {
+      return <Link href={this.props.href}>{anchor}</Link>;
```
smcmurtry, line 68: I like how you did this!

smcmurtry (approved):
Nice work 🦑

sastels (approved):
Good job 🐝

Issue #1223, opened 2018-10-02
Upgrading to react-i18next v8.0.6 breaks app


smcmurtry:
We'll have to fix this at some point unless we want to be stuck on our current version.

brdunfield:
![](https://github.trello.services/images/mini-trello-icon.png) [Technical Debt Sprint 3](https://trello.com/c/Q3bSJUnL/79-technical-debt-sprint-3)

brdunfield:
closed with #1436

PR #1385, opened 2018-10-31
Security update

brdunfield:
#1381


sastels:
I think that the `react-i18next` major upgrade is still breaking the app, as mentioned in https://github.com/cds-snc/vac-benefits-directory/issues/1223. We should probably fix this :/

brdunfield:
Do you want me to revert just the breaking dependency on this PR so the other security updates go through? And move #1223 to the current sprint?

brdunfield:
Should be working now!

sastels (approved):
Looks good! 🦑

PR #1389, opened 2018-11-01
Benefit titles are now h2 for accessibility

### LINKED ISSUE #1356: Benefit titles should be marked as headings
From Level Access:

Avoid the use of implicit headings.

[Issue]
The page contains textual content that visually appears to be a heading, but is not marked up as such.

[Impact]
Assistive technology users may have difficulty determining the page structure or efficiently navigating the page, as many assistive technology users use headings to navigate applications.

[Expected Behavior]
Ensure that all visually apparent headings are marked up as such. Developers should use heading mark up on the benefit titles to allow for more efficient navigation and more defined page structure.

SupeDeDupe:
Closes #1356

### DIFF
--- a/__tests__/components/typography/benefit_title_test.js
+++ b/__tests__/components/typography/benefit_title_test.js
@@ -1,0 +1,23 @@
+import React from "react";
+import { mount } from "enzyme";
+import BenefitTitle from "../../../components/typography/benefit_title";
+const { axe, toHaveNoViolations } = require("jest-axe");
+expect.extend(toHaveNoViolations);
+
+describe("BenefitTitle", () => {
+  let props;
+  beforeEach(() => {
+    props = {
+      children: "header"
+    };
+  });
+
+  it("passes axe tests", async () => {
+    let html = mount(<BenefitTitle {...props} />).html();
+    expect(await axe(html)).toHaveNoViolations();
+  });
+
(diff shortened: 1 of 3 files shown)


brdunfield:
H2 is the same level as the page title(s). Should the cards perhaps be H3, or promote the page titles to an H1 to better indicate the hierarchy of the page? Should individual benefit cards have the same hierarchy as the overall page title?

SupeDeDupe:
I feel like we should maintain the hierarchy, but I'm not sure where you mean it's the same level. Can you give me a content-specific example?

brdunfield:
screen shot 2018-11-01 at 2 42 57 pm
For example, on this page, "Showing all benefits" is an H2, as is the card title "Disability Award".

SupeDeDupe:
Oh yeah, I see what you're saying. I don't see why there would be a problem in making it h3, so I'll try doing that.

SupeDeDupe:
On second thought, Should there be an h1 on this page? Can't seem to find one, and that seems odd.

smcmurtry:
Thanks Sarah, I think that's a good solution!

brdunfield:
I noticed that too Sarah - should the page title maybe be promoted to an H1?

SupeDeDupe:
I vote to change the page title to h1.

Issue #1390, opened 2018-11-02
Button and Link styling


brdunfield:
Potentially an accessibility issue: The app currently has <button> elements styled as links, and <a> elements styled as buttons.

A conversation should be had between dev, design, and ideally looped in with Julianna to decide where to proceed on this issue.

PR #1391, opened 2018-11-02
Render a <button> if useLink is not present. 

### LINKED ISSUE #1388: Refactor / Split HeaderButton component in two
Some HeaderButtons should be anchors, others should be buttons. We need a second component (HeaderLink ?) so the correct HTML tag can be used based on the function of the component.

brdunfield:
Add useLink to HeaderButtons in the benefit cards that are actually links.

Resolves #1388 but not #1390 - one step along this process.


sastels (approved):
Good job! 🦑

PR #1396, opened 2018-11-05
Added alternative text to the important info icon

### LINKED ISSUE #1348: Add alternative text to the ! icon
Currently users using screen readers will not have any info about what the icons mean. Based on discussion with Julianna, we should use alt-text to fix this. We should test with a screen reader (voiceover on osx) to verify that it is working.

smcmurtry:
Closes #1348 .

Tested with voiceover on osx, it successfully read the alternative text for the icon.

Also I deleted some unneeded files.


sastels (approved):
LGTM 🦑

PR #1400, opened 2018-11-06
Making headers more consistent

### LINKED ISSUE #1398: Take <h1>, <h2>, ... tags out of typography components and define them on the page
use:

`<h1><header1 /></h1>`

on the page, where header is a styled `<span>`, rather than:

`<header1 />`

where header contains an `<h1>`.

Relevant discussion from Slack:

https://webdesign.tutsplus.com/articles/the-importance-of-heading-levels-for-assistive-technology--cms-31753

"The real world may call for a fifth level heading that needs to look like it is a second-level heading, or something entirely new. In these situations it’s better to preserve the underlying logical page structure and modify the heading’s visuals instead".

That tells me it's less important for an `<h2>` to look the same everywhere. And looking on gov.uk, I just noticed they have `<h2>`s on the same page that are styled very differently, from 16px font to 36px.

So we should just specify the heading level case by case, and not build components that are the same heading level everywhere.

smcmurtry:
Closes #1398

Still more work to be done to add headers in the other places in the app. Will make another issue for that.


brdunfield:
I feel like we could simplify this a lot more than it currently is. The only difference between the header components (that I can see) is styling, so I feel like they could be combined with a prop to indicate which style you want to apply, or which heading html you want to wrap. I'm not sure how I feel about putting the heading html out of a "heading" component.

smcmurtry:
@brdunfield I think that's a good idea to combine the header components. Regarding putting the `<h1>`, `<h2>` tags outside of the header components, that was the result of a discussion with Julianna on vac-devs. If you have a different proposal you could join the conversation there.

brdunfield (approved):
Awesome! 🦑

PR #1408, opened 2018-11-07
Fix back button not working in Guided Experience

### LINKED ISSUE #1404: Back on first GE doesn't work
Doesn't go back to landing page anymore. back to previous GE pages works.

brdunfield:
There was a bug in the router where the "onRouteChangeStart" function was globally overriden, but the override only applies to the guided experience. Also changed the initial back button to a link instead of a button.

Closes #1404 

This does however introduce a new bug where the back button must be clicked twice on the first GE page.. the first click clears the radio button and the second click navigates back. I've investigated this a bit but haven't been able to figure out why..

### DIFF
--- a/components/guided_experience.js
+++ b/components/guided_experience.js
@@ -108,7 +108,7 @@ export class GuidedExperience extends Component {
       <Container id="guidedExperience">
         <HeaderButton
           id="prevButton"
-          disableRipple
+          useLink={prevSection === "index"}
           href={
             prevSection === "index"
               ? "/index?lng=" + t("current-language-code")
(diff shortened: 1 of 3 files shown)


smcmurtry (approved):
Hmm I'm not sure why that's happening either. Let's make an issue for it.

smcmurtry:
Actually it looks like the bug is due to us setting the url incorrectly in the componentDidUpdate in guided.js

sastels:
yes I remember this sort of bug appearing when I was poking at it a while ago. I think I was trying to use `Link`s for those buttons instead of anchors? I remember being very unhappy.

PR #1409, opened 2018-11-07
AnchorLink component

### LINKED ISSUE #1407: Fix focus styles on phone number on map page

### LINKED ISSUE #1411: Phone number and email links don't work on favourites page

smcmurtry:
Changed the guided experience link component to a more general anchor link component.

Used this component on the map page so the phone number has the yellow focus style.

Replaced some header buttons on favourites since they were broken.

Also implemented a design change for contact info on the map page.

Closes #1407 
Closes #1411


brdunfield:
How does this component relate to our ongoing process of splitting HeaderButton into button and link components? Are these anchors just <a> elements *not* wrapped in <Link>s?

I'm wondering if as part of the HeaderButton refactor that the <Link> wrapping logic should be moved into this component - as the buttons in HeaderButton are never wrapped in the Link.

Essentially I'm proposing we open a new issue where we eventually make HeaderButton component render <button> elements only, and have this component render anchors with / without the <Link> wrapper as needed? Assuming I'm understanding the function of this new component correctly..

None of the above relates to this particular PR, just want to pose the question!

components/typography/anchor_link.js
```diff
@@ -0,0 +1,48 @@
...
+      outline: 3px solid ${globalTheme.colour.focusColour};
+    }
+  `;
+
+  render() {
+    const { className, children } = this.props;
+
+    return (
+      <a
+        className={className ? cx(this.style, className) : this.style}
```
sastels: I think you can just do `cx(this.style, className)`, ie`cx(this.style, undefined) == this.style`

components/typography/anchor_link.js
```diff
@@ -0,0 +1,48 @@
+import React, { Component } from "react";
+import PropTypes from "prop-types";
+import { globalTheme } from "../../theme";
+import { cx, css } from "react-emotion";
+
+class AnchorLink extends Component {
```
sastels, line 6: Should this just be called `Anchor` instead? I assumed by the name that it was going to use a next `Link`

sastels (approved):
minor suggestions, LGTM (Bethany has a good question we should figure out, though)

components/typography/anchor_link.js
smcmurtry: Bethany and I discussed IRL and we'll implement the `<Link>` wrapper in this component and remove that functionality from the HeaderButton (so that will always return a button...)

PR #1415, opened 2018-11-07
added focus outline to search button and clear button, added disabled…

### LINKED ISSUE #1410: Fix focus style on search button
Where the button is clickable it should be the yellow focus style, where it is not clickable it should not have a focus style.

smcmurtry:
… prop to the button when disabled

Closes #1410

### DIFF
--- a/components/search_box.js
+++ b/components/search_box.js
@@ -53,7 +53,8 @@ const ClearButton = styled("button")({
   backgroundRepeat: "no-repeat",
   backgroundPosition: "2px 50%",
   ":focus": {
-    outlineOffset: 0
+    outlineOffset: 0,
+    outline: `3px solid ` + globalTheme.colour.focusColour
   },
   ":hover": {
     color: globalTheme.colour.black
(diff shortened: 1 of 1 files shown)


components/search_box.js
```diff
@@ -53,7 +53,8 @@ const ClearButton = styled("button")({
   backgroundRepeat: "no-repeat",
   backgroundPosition: "2px 50%",
   ":focus": {
-    outlineOffset: 0
+    outlineOffset: 0,
+    outline: `3px solid ` + globalTheme.colour.focusColour
```
sastels, line 57: We could move the entire  \`3px solid\` + globalTheme.colour.focusColour into a `globalTheme.focusOutline` if we wanted.

sastels (approved):
minor comment, 🦑

components/search_box.js
smcmurtry: that's a good idea. want to make an issue for it? :)

PR #1424, opened 2018-11-09
add useLink to favourite page print button

sastels:
fixes #1423

PR #1425, opened 2018-11-09
test new airtable read key

sastels:
empty PR to test that using the vac-cds airtable key for reading works both in Heroku and Azure


sastels:
Works with Heroku 🤷‍♂️

sastels:
already tested by merging #1424 🎉

PR #1436, opened 2018-11-14
Replacing react-i18next

brdunfield:
Throwing this up as a WIP for feedback / comments on @smcmurtry 's suggestion for #1223 .

I'm not very happy with this whole situation. I'm making progress here but there's still a fair bit to do, I think, before it's at a state where it can fully replace react-i18next.

Currently there are 2 HOCs, one for the server, and one for the client (because the server _document.js page can't use redux's connect function?). This is something I'd like to fix in the future since the two only differ in a couple lines for accessing the translation data.

Current State: the HOCs work but are hard-coded to return English only, and language switching is completely broken. Any feedback or comments on how to best get / make use of the current language would be appreciated as that's the next major hurdle.

More general feedback on this entire issue and what we should be doing in situations like these would also be great. The entire situation is very frustrating - upgrading the dependency was shaping up to be a ton of work, but so is this issue / PR.

@sastels @maxneuvians @SupeDeDupe @leonL


maxneuvians:
@brdunfield - here would be my recommendation. Ditch the HoC, they may not be needed. If I recall correctly all the translation strings are loaded into memory and can also therefore be loaded into redux. It looks like this is already happening here: https://github.com/cds-snc/vac-benefits-directory/blob/master/store.js#L78-L82

What you would then do is also create a redux state for the language. This is happening a bit already in these selectors: https://github.com/cds-snc/vac-benefits-directory/blob/master/selectors/benefits.js#L6

All you need to do is create a new selector like the ones in here: https://github.com/cds-snc/vac-benefits-directory/tree/master/selectors

These selectors can be treated like filtered versions of redux state. See here: https://github.com/cds-snc/vac-benefits-directory/blob/e5cfde3953b81e9b5a943872d34b8956f5a07e7f/pages/benefits-directory.js#L8 and here: https://github.com/cds-snc/vac-benefits-directory/blob/e5cfde3953b81e9b5a943872d34b8956f5a07e7f/pages/benefits-directory.js#L94

What you in the end have is a property in the component that represents a key-map of string translations. So when you have a change in language the selector will change that map to the right language and because the keys are the same in both English and French maps, they should just translate. Let me know if that makes sense, if not we can hash it out in person. Also maybe you have tried this approach and it did not work as expected, in which case ignore what I said.

brdunfield:
I have removed the server-side HOC (see #1441 ) so now there is only a single HOC for the app again 🎉 

Note: a yarn install will be required to run this branch

smcmurtry (approved):
I tested setting my browser language to french and the app automatically loaded the french content.

I tested performance (using lighthouse) and it seems a bit better. Page data transferred went from 453 KB to 424 KB. 

I tested changing content in airtable and refreshing the cache, and the page picks up the changes correctly. 

I've looked over the code and don't see any issues.

Great job @brdunfield!!!! 🦑

Issue #1421, opened 2018-11-09
First back button in Guided Experience needs 2 clicks


brdunfield:
If a radio item is set, for this screen only, navigating back takes too clicks.

See #1408 for more info.

brdunfield:
I believe this bug is still present, but it is very difficult to test as the initial back button currently doesn't go anywhere, but will eventually link back to the VAC homepage.

smcmurtry:
I switched the link to the VAC home page in airtable and the button is now working as expected with 1 click.

PR #1440, opened 2018-11-16
Radio button swap

### LINKED ISSUE #1362: Swap out material-ui radio button component for gov.uk one
This will make it easier to have a consistent focus style across all components. And it should also fix the broken keyboard navigation that exists for our current radio component.

https://github.com/UKHomeOffice/govuk-react

smcmurtry:
Closes #1362

Keyboard nav isn't working quite yet, but that can be solved in another issue.

### DIFF
--- a/__tests__/components/radio_selector_test.js
+++ b/__tests__/components/radio_selector_test.js
@@ -35,16 +35,14 @@ describe("RadioSelector", () => {
     expect(await axe(html)).toHaveNoViolations();
   });
 
-  it("has 3 FormControlLabels", () => {
-    expect(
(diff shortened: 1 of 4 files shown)


smcmurtry:
hmm, not working quite yet

SupeDeDupe:
The back button won't allow the user to navigate back to the home page.

smcmurtry:
@SupeDeDupe if you check again, the behaviour is the same as in master, so this is the existing bug #1421 and not related to this PR.

brdunfield (approved):
Looks a lot simpler than the old implementation - nice job! 🦑

PR #1443, opened 2018-11-19
Data-validation renders benefit names instead of IDs

### LINKED ISSUE #1386: change benefit ids in data validation page to names
in the "benefits without needs" row we show the ids of the benefits without needs. we should show the names of the benefits instead of the ids.

brdunfield:
Closes #1386


smcmurtry (approved):
Looks good! 🦑

PR #1448, opened 2018-11-20
Sorting by search score

### LINKED ISSUE #1445: Sort benefits by lunr score
If the user has entered text in the search box, sort the matching benefits by the score that lunr returns.

smcmurtry:
Closes #1445 

I wanted to test the sort order, but the existing hand-coded indexes only returned a match for 1 benefit and I couldn't figure out how to add to the index manually. So I added new fixtures for the lunr-generated index files from the app. But to use those, I needed to add the real list of benefits, eligibility paths, etc. So now we have a bunch of new fixtures with names ending in `_complex`.

Let me know what you think. Maybe I should change them to be in a `real_app_data` subfolder in the fixtures folder.


sastels (approved):
Wow, great job!! 👏 

I don't see anything wrong with having a special set of more complicated fixtures to use to test a more complicated feature. You could perhaps pare down those fixtures a bit (maybe delete some fields the search doesn't care about, pare down the number of elements in the fixtures to just those needed for a good test) but that would probably take quite a bit of time with not a great deal of benefit.

Issue #1452, opened 2018-11-22
Implement Share Button and modal


brdunfield:
Copy link to clipboard.

brdunfield:
![](https://github.trello.services/images/mini-trello-icon.png) [Users are able to copy the URL and share to their networks](https://trello.com/c/q8cZ0odM/102-users-are-able-to-copy-the-url-and-share-to-their-networks)

smcmurtry:
Closed by #1453

PR #1453, opened 2018-11-22
Share Button with link copy functionality

brdunfield:
#1452

New ShareModal component. Fairly specific, if we wanted to make other modals in the future we would need to generalize this component.


brdunfield:
[Users are able to copy the URL and share to their networks](https://trello.com/c/q8cZ0odM/102-users-are-able-to-copy-the-url-and-share-to-their-networks)

brdunfield:
[Users are able to copy the URL and share to their networks](https://trello.com/c/q8cZ0odM/102-users-are-able-to-copy-the-url-and-share-to-their-networks)

brdunfield:
I need to write some tests for this before I take the [WIP] off the PR, but it should be otherwise functional in the app / deployment if anyone is interested or wants to work on other aspects of it!

It’s a bare-bones-ish implementation, so I expect future issues / functions to be added. In particular I would like to add some kind of confirmation that the link was successfully copied, but a chat with @smcmurtry  hinted that a similar function may be implemented elsewhere soon, so we can probably just duplicate that work when it happens.

smcmurtry (approved):
Great work! Let's add an issue for the styling bug.

brdunfield:
There are some outstanding bugs that will be made into issues, including:
* Weird styling bug sometimes along header of app - happens when the app with a query string is loaded, but does not happen if page is navigated to from home
* warning about App element being undefined in the share_modal tests. This is an artifact of the react-modal dependency component.
* code coverage of share_modal tests can be improved.

PR #1454, opened 2018-11-22
Remove sortby dropdown

### LINKED ISSUE #1444: Remove the sort by dropdown
We should always sort by the priority column in airtable, then alphabetically, unless the user has entered text in the search box. Because we will always be setting this programmatically, we can remove the sort by dropdown from the app.

SupeDeDupe:
Closes #1444


smcmurtry:
Looks good! I noticed that searching the code for "sortBy" turns up a lot of other places where it needs to be removed: url parsing logic, redux, other components, etc. You could either fix it here or make another issue and do that as a separate PR.

SupeDeDupe:
Ok it seems to get pretty complicated here and I think it'll require a much deeper look into all the places where it's being to understand how it's being used, and then how it would be safely removed. Just created a new issue for it (#1456).

smcmurtry (approved):
Great work! 🦑

PR #1459, opened 2018-11-23
removed all references to sortby, made priority then alphabetical the default sorting

### LINKED ISSUE #1456: Removed all parts of code with "sortBy"
The sortBy dropdown was removed in a previous issue.  Searching the code for "sortBy" turns up a lot of other places where it needs to be removed: url parsing logic, redux, other components, etc. Also to get rid of the warnings associated with this dropdown removal.

smcmurtry:
Closes #1456 

Also, the warnings to do with missing sortby props are gone.


step_by_step.txt
```diff
@@ -0,0 +1,12 @@
+1. Save the benefits you're interested in
```
sastels: Not sure what exactly this file is for?

sastels (approved):
Nothing wrong with getting rid of code! 🦑

step_by_step.txt
smcmurtry: whoops! that made it in by accident.

PR #1472, opened 2018-11-26
Fixed layout bug

### LINKED ISSUE #1460: Header bar render issue
happens when the app with a query string is loaded, but does not happen if page is navigated to from home page.

smcmurtry:
Closes #1460

Note: Moving the share modal component is what fixed the bug, but I also changed the xs column widths to give a bit more room to the 2 header buttons on the right


brdunfield (approved):
👍 🦑

PR #1484, opened 2018-11-28
Added Skip button, which is an alternate styled HeaderButton

brdunfield:
I'm not sure how I feel about this approach.

I still think our approach to buttons / links needs some refactoring. I didn't want to make yet another basic button component that's only unique feature is its style, but I'm not sure that HeaderButton is in the right state to handle this kind of customization. Some refactoring might be needed..

Thoughts?

### DIFF
--- a/components/guided_experience.js
+++ b/components/guided_experience.js
@@ -151,7 +151,7 @@ export class GuidedExperience extends Component {
               {this.props.children}
             </Grid>
 
-            <Grid item xs={4}>
+            <Grid item xs={12}>
               <Button
                 id="nextButton"
                 arrow={true}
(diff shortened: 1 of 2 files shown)


smcmurtry:
Nice! I think you need less white space between the next and skip buttons (the design is about 40px). And I noticed all the font sizes increased on that page to 24px, but maybe that should be another PR.

I still think we should split a `HeaderLink` component out of `HeaderButton`, which I think would simplify things. Maybe I will do that after this PR merged.

brdunfield:
@smcmurtry I don't think the font-size was affected by this PR, so agree it should be a separate issue.

I fixed the spacing and made the "Skip question" read a translation from AirTable.

smcmurtry (approved):
Nice work! 🦑

PR #1490, opened 2018-11-28
Upgrade share copy function to use navigator.clipboard

### LINKED ISSUE #1475: Share link modal doesn’t copy link on mobile
At least, not on my iPhone XS

brdunfield:
Resolves #1475 
document.execCommand is deprecated - this is probably why #1475 occurred on a newer device.

### DIFF
--- a/components/share_modal.js
+++ b/components/share_modal.js
@@ -115,8 +115,10 @@ class ShareModal extends Component {
   constructor(props) {
     super(props);
     this.state = {
-      url: ""
+      url: "",
+      statusMessage: ""
     };
+    this.copyText = this.copyText.bind(this);
   }
   componentDidMount() {
     this.setState({ url: window.location.href });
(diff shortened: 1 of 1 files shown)


smcmurtry:
@sastels mind testing this with your iphone? :)

brdunfield:
https://stackoverflow.com/questions/34045777/copy-to-clipboard-using-javascript-in-ios

It's not a navigator.clipboard issue (although we should still be using that). It's an iOS issue (>10.0). I've been fighting with it for a few hours. I've gotten it so it'll select the text as indicated in the stackoverflow, but document.execCommand("copy") still doesn't seem to be doing anything.

brdunfield:
So when testing in browserstack, I finally realized the "device" clipboard is not the same as my computer clipboard. If I use the on-device "paste" button (into the url field for example), I can verify that the link was copied.

@sastels I might still appreciate a quick test on a real device though!

brdunfield:
Also added a really basic implementation of #1497 to help with the debugging. Does not fully resolve this issue, but thought I should mention it.

smcmurtry:
I tested on browserstack too and the copy does seem to be working now! However, when the "link copied" text shows up it seems to cause the device to magnify the page:

Before:
screen shot 2018-11-29 at 11 38 27 am

After:
screen shot 2018-11-29 at 11 37 41 am

So maybe we could leave that text out for now.

sastels (approved):
works for me 🎉 ! maybe deal with that magnification issue if possible.
🦑 🦑

PR #1492, opened 2018-11-28
Split out new header link component

### LINKED ISSUE #1412: Remove the anchor/link functionality from HeaderButton, implement in new AnchorLink component

smcmurtry:
Closes #1412 

We already have a component named `AnchorLink` so I called it `HeaderLink`. We now have 2 different components that are styled the same way, but at least now it is obvious whether we're using a  `Link` or a `button`.


brdunfield (approved):
Looks good to me. I'm wondering if there's some way we can abstract the style out of the components - i.e. only have the style code written once and both components import it? Something for another issue though.


🦑!

PR #1494, opened 2018-11-28
fixed the bug where the share modal was not getting the most current url

### LINKED ISSUE #1491: Share button - url not updated until refresh occurs
To reproduce:

1. visit https://cdsvacpoc.azurewebsites.net/benefits-directory
2. select some radio buttons
3. click share
4. the share modal url doesn't match the current browser url

However, if you refresh the page, the urls do match.

To fix this, Max suggests getting the url from the selector.

smcmurtry:
Closes #1491 

This turned out to be fairly easy since the current url is passed as a prop to each page, so when it changes that will trigger the modal to re-render.

### DIFF
--- a/__tests__/components/share_modal_test.js
+++ b/__tests__/components/share_modal_test.js
@@ -21,6 +21,7 @@ describe("ShareModal", () => {
       isOpen: true,
       closeModal: jest.fn(),
       t: () => "en",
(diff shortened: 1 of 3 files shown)


brdunfield (approved):
💯   🦑

Issue #1507, opened 2018-11-30
Create a breadcrumb nav bar component


smcmurtry:
For now, all it needs to do is have "Home / Current page name", with the current page name passed in as a prop. 

To start we could just put it one page (maybe the map page?).

<img width="970" alt="screen shot 2018-11-30 at 12 10 06 pm" src="https://user-images.githubusercontent.com/5498428/49303903-f3f62000-f498-11e8-9e1b-94ef4b092ff8.png">

Design: https://app.zeplin.io/project/5acbd659dc553ec84805afd4/screen/5c006921b6aa8378fa8e80c8

PR #1514, opened 2018-11-30
Implement basic breadcrumbs on maps page

brdunfield:
Implements #1507


smcmurtry (approved):
Looks great! Minor suggestions for another PR: background is greyish, Home needs translating, and the home link could be hardcoded rather than a prop since it should always be the same.

brdunfield:
I left the background white for now until the rest of the pages get re-done to have the white background. I'll make a new PR for these changes.

Issue #1531, opened 2018-12-04
Remove share button from favourites page


smcmurtry:
The share feature doesn't work on this page because the favourites list is kept in cookies, not the url. So let's remove it for now.

brdunfield:
Closed by #1533

PR #1533, opened 2018-12-05
Add breadcrumbs to benefits-directory and favourites page

### LINKED ISSUE #1522: Add breadcrumb navigation to benefits-directory, favourites
Also change the background colour to be more grey: https://app.zeplin.io/project/5acbd659dc553ec84805afd4/screen/5c006921d385f82c6d6bb073

brdunfield:
Closes #1522 and also #1531

### DIFF
--- a/components/BB.js
+++ b/components/BB.js
@@ -75,6 +75,7 @@ export class BB extends Component {
           <BreadCrumbs
             t={t}
             breadcrumbs={[]}
+            url={url}
             pageTitle={t("ge.Find benefits and services")}
           />
           <Container>
--- a/components/BB.js
+++ b/components/BB.js
@@ -8,7 +8,7 @@ import Bookmark from "./icons/Bookmark";
 import ProfileNeedsSelector from "./profile_needs_selector";
 import ProfileNeedsSelectorMobile from "./profile_needs_selector_mobile";
 import { connect } from "react-redux";
-import { getFavouritesUrl, getPrintUrl } from "../selectors/urls";
+import { getFavouritesUrl, getPrintUrl, getHomeUrl } from "../selectors/urls";
 import { css } from "react-emotion";
 import Container from "../components/container";
 import HeaderButton from "./header_button";
@@ -63,7 +63,7 @@ export class BB extends Component {
   }
 
   render() {
-    const { t, url, store } = this.props; // eslint-disable-line no-unused-vars
+    const { t, url, store, homeUrl } = this.props; // eslint-disable-line no-unused-vars
 
     return (
       <div
@@ -75,7 +75,7 @@ export class BB extends Component {
           <BreadCrumbs
             t={t}
             breadcrumbs={[]}
-            url={url}
+            homeUrl={homeUrl}
             pageTitle={t("ge.Find benefits and services")}
           />
           <Container>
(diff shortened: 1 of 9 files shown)


components/favourites.js
```diff
@@ -90,42 +82,29 @@ export class Favourites extends Component {
       this.props.favouriteBenefits
     );
 
+    const breadcrumbs = [
+      { url: "/benefits-directory", name: t("titles.all_benefits") }
```
smcmurtry: if you make some radio selections on /benefits-directory, click the button to go to the favourites page, and then click the "all benefits" breadcrumb, the url query parameters for the selections are now gone. I think if you use the `common/mutate_url` function, that will fix the issue.

components/BB.js
```diff
@@ -71,7 +72,12 @@ export class BB extends Component {
         ref={el => (this.componentRef = el)}
       >
         <div className={topMatter}>
-          <Container className={topPadding}>
+          <BreadCrumbs
+            t={t}
+            breadcrumbs={[]}
+            pageTitle={t("titles.all_benefits")}
```
smcmurtry: I think the breadcrumb text for benefits-directory should be "Find benefits and services":
https://app.zeplin.io/project/5acbd659dc553ec84805afd4/screen/5c006920d385f82c6d6bafaa

smcmurtry: this just needs to be updated to your new title: `t("ge.Find benefits and services")`

brdunfield: Good catch - should be fixed now.

components/breadcrumbs.js
```diff
@@ -25,21 +26,36 @@ const separator = css`
...
+    const { breadcrumbs } = this.props;
     return (
-      <div className={whiteBanner}>
+      <div className={greyBanner}>
         <Container>
           <div className={breadCrumbStyle}>
-            <a href={this.props.url} className={urlStyle}>
-              <HomeIcon className={iconStyle} /> Home
-            </a>
+            <HeaderLink id="homeButton" href={"/"} className={urlStyle}>
```
smcmurtry: One more bug: if you're on /benefits-directory, you make some selections, then click the Home breadcrumb, the url params are cleared and you stay on the same page without going to home. 

I think this would be fixed with `href={mutateUrl(url, "/")}`

components/favourites.js
```diff
@@ -90,42 +82,32 @@ export class Favourites extends Component {
...
+      }
+    ];
+
     return (
       <div className={outerDiv}>
+        <BreadCrumbs
+          t={t}
+          url={url}
+          breadcrumbs={breadcrumbs}
+          pageTitle={t("index.your_saved_benefits")}
```
SupeDeDupe, line 98: I was thinking maybe the pageTitle should reference the same record in AirTable that the actual page title is referencing, in this case favourites.saved_benefits, so if an admin ever changes the page title, the breadcrumbs will remain in sync with the updated page titles. But this could also just be added in a separate issue.

smcmurtry:
One last bug, to reproduce:
1. go to https://vac-poc-pr-base-pr-1533.herokuapp.com/benefits-directory?lng=en
2. make some radio selections
3. click home
4. the home button won't go back to home and will change the radio selection

If you refresh the page between steps 2 and 3, the home button works as expected.

So I think the issue is that the url prop getting passed to mutateUrl is old. Maybe it would work if we pass redux state in and make the url using that?

components/favourites.js
brdunfield: Yes I think this is a separate issue, because the title on the directory is also not the same as the one being used in the breadcrumbs.

smcmurtry (approved):
Great work Bethany! This turned out to be a tough one. 🦑

brdunfield:
[User is overwhelmed viewing the directory for the first time after going through the guided experience](https://trello.com/c/MM7hS6SQ/113-user-is-overwhelmed-viewing-the-directory-for-the-first-time-after-going-through-the-guided-experience)

PR #1535, opened 2018-12-05
Making the filter default collapsed

### LINKED ISSUE #1508: Change the Filters dropdown to always be the mobile expandable component
Also the header now has blue text: https://app.zeplin.io/project/5acbd659dc553ec84805afd4/screen/5c006920d385f82c6d6bafaa

screen shot 2018-12-05 at 3 16 18 pm

smcmurtry:
Closes #1508


SupeDeDupe (approved):
The reset button is small in non-mobile and should probably be bigger, but maybe this could be changed in a separate issue that also involves moving the reset button. It used to be at the top and there doesn't seem to be a current design available for its position in the filter dropdown.

brdunfield:
[User is overwhelmed viewing the directory for the first time after going through the guided experience](https://trello.com/c/MM7hS6SQ/113-user-is-overwhelmed-viewing-the-directory-for-the-first-time-after-going-through-the-guided-experience)

PR #1538, opened 2018-12-06
Next and back buttons scroll to the top of the new page

### LINKED ISSUE #1528: App should jump to top of page when moving between pages
When moving between pages of the app, the app remains scrolled to where you were on the previous page. Expected it would jump to the top. 

Repro:
1. On the first page of the app, scroll down, then hit next: 
image
2. Observe that the next page is scrolled down too: 
image

This is very obvious when it's the last page:
image
hit results:
image

smcmurtry:
Closes #1528

### DIFF
--- a/__tests__/components/guided_experience_test.js
+++ b/__tests__/components/guided_experience_test.js
@@ -30,7 +30,8 @@ describe("GuidedExperience", () => {
     }
     return _shallowGuidedExperience;
   };
-  Router.push = jest.fn();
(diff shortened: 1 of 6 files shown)


SupeDeDupe:
I think if you delete the test for getNextSection, that'll fix the error.

brdunfield (approved):
LGTM! 🦑

courtneycl (approved):

PR #1575, opened 2018-12-10
Upgrade emotion

### LINKED ISSUE #1526: update emotion and related packages
A substantial update to `emotion` and related packages includes significant breaking changes. Hopefully the migration steps suggested in https://emotion.sh/docs/migrating-to-emotion-10 will make the pain less.

sastels:
resolves #1526 

Partially migrate the code to emotion 10, as described in https://emotion.sh/docs/migrating-to-emotion-10. Essentially, it all works, but is still using the `className` prop rather than the new `css`. Code can be changed to use `css` rather than `className` as we go. I'll probably do a bunch after this merges to start the process.

The app has changed so much since I worked on it that I *think* everything is still being styled correctly, but it's probably a good idea to give it a good look over.


yarn.lock
```diff
@@ -5,7 +5,6 @@
 "@babel/cli@^7.2.0":
   version "7.2.0"
   resolved "https://registry.yarnpkg.com/@babel/cli/-/cli-7.2.0.tgz#505ed8d351daee6a88918da02c046c18c8c5a24f"
-  integrity sha512-FLteTkEoony0DX8NbnT51CmwmLBzINdlXmiJCSqCLmqWCDA/xk8EITPWqwDnVLbuK0bsZONt/grqHnQzQ15j0Q==
```
sastels, line 8: not sure why all these lines got deleted? `yarn` works in mysterious ways...

brdunfield (approved):
No firebreak needed! 👍 🦑

smcmurtry:
I took a look and I couldn't see any styling changes introduced by this PR. Looks good!

PR #1576, opened 2018-12-10
Created test file for next steps

### LINKED ISSUE #1548: Create test file for Next Steps

SupeDeDupe:
Closes #1548


smcmurtry:
Can you add some more tests? Ideally the tests should cover the main functionality for the component. For next steps that is currently just rendering static content. So you could test:
1. that the bullet list is rendered
2. that the 2 cards are rendered

__tests__/components/next_steps_test.js
```diff
@@ -0,0 +1,29 @@
...
+  beforeEach(() => {
+    props = {
+      t: translate,
+      mapUrl: { query: {}, route: "/map" }
+    };
+
+    mockStore = configureStore();
+    reduxState = {};
+    props.store = mockStore(reduxState);
+    props.reduxState = reduxState;
```
smcmurtry, line 19: I think you can delete lines 19 and 21 and this should still work

SupeDeDupe:
Not sure how to fix this warning:
screen shot 2018-12-11 at 10 32 08 am

__tests__/components/next_steps_test.js
SupeDeDupe: I think I removed them

__tests__/components/next_steps_test.js
```diff
@@ -0,0 +1,71 @@
...
+const { axe, toHaveNoViolations } = require("jest-axe");
+expect.extend(toHaveNoViolations);
+
+describe("NextSteps", () => {
+  let props;
+  let reduxState;
+  beforeEach(() => {
+    props = {
+      t: translate,
+      mapUrl: { query: {}, route: "/map" }
```
smcmurtry: This should actually just be a string - it's different from the url prop that is passed to each page, it's the result of the `getMapUrl` function in selectors/urls.js. That should fix your warning.

smcmurtry (approved):
Good work! 🦑

__tests__/components/next_steps_test.js
SupeDeDupe: But how do I make it a string and still the thing I want?

smcmurtry: `mapUrl: "/map"`

SupeDeDupe: Ok but then what do I put in the test for it on line 69? I think if I write "/map" then does the test still make sense??

smcmurtry:
[Users are given options for what the next step is in their current stage](https://trello.com/c/rHNRWLYF/2-users-are-given-options-for-what-the-next-step-is-in-their-current-stage)

PR #1587, opened 2018-12-12
Fix bug in search headers

### LINKED ISSUE #1586: Making selections causes directory title to be "{x} Search Results"

brdunfield:
Closes #1586


smcmurtry (approved):
Looks good! Only issue is the case with 1 benefit but that can be a separate PR

PR #1589, opened 2018-12-12
Fix SSR

### LINKED ISSUE #1582: janky beta banner
The upgrade to emotion 10 seems to have had the effect that the SSR of the Beta banner isn't quite the same as the client side rendering 😞 The effect is that on load the banner is initially a few pixels shorter, then pops in at full size. Text in the banner is correct throughout.

sastels:
fixes #1582 

Problem was that gcui is based on emotion 9, this didn't play well with emotion 10 and SSR.

Removed our dependency on gcui by copying `PhaseBadge` and `WordMark` into the VAC codebase (and changing both to emotion 10)


components/phase_badge.js
```diff
@@ -0,0 +1,26 @@
+import React from "react";
+import PropTypes from "prop-types";
+import styled from "@emotion/styled";
+import { globalTheme } from "../theme";
+// import { mediaQuery } from '../utils'
```
smcmurtry: 🔥

smcmurtry (approved):
Looks fixed, thank you!! 🦑

PR #1593, opened 2018-12-12
Search also shows all benefits results

### LINKED ISSUE #1568: Search: Add results from all benefits
Exclude results that were already displayed in "results from your selections"

Will also need to update the number of results in the page title

brdunfield:
Closes #1568


brdunfield:
[Users are able to differentiate between the filtered benefits and all benefits while searching](https://trello.com/c/ZjrlilF0/127-users-are-able-to-differentiate-between-the-filtered-benefits-and-all-benefits-while-searching)

smcmurtry (approved):
The behaviour in the app is what I would expect, and the code looks good. I like that you split out the search into its own function. 🦑

PR #1595, opened 2018-12-12
 Made a function in utils/common to give a string with the number of matching benefits

### LINKED ISSUE #1592: Start displaying "one benefit or service matches your selections" text
Change "x benefits and services match your selections" text on benefits-directory and summary page so that it's not plural if there's only 1 benefit.

smcmurtry:
Closes #1592


brdunfield (approved):
LGTM! 🦑 

There will probably be some conflicts in #1593 as a result of this merge.

Issue #1441, opened 2018-11-19
Language Switching does not change some functions in the app


brdunfield:
Notably:
* html "lang" attribute does not change
* map is loaded from initial language and does not change (so if app is first loaded as eng, the google map will never have french UI elements unless the app is reloaded with fr as the initial language)

Both of these are set in _document.js

This is tangentially related to #1436 , in that the i18n that was previously used in _document was doing nothing, so it has been removed in that PR.

brdunfield:
Lang attribute now changes on language change, and map is no longer part of the app. Closing!

PR #1604, opened 2018-12-13
use react-helmet to set page language

sastels:
deals with part of #1441. Sets page language when language button pushed. Doesn't change google map script.


smcmurtry (approved):
LGTM!

PR #1607, opened 2018-12-13
Security update sprint 6

SupeDeDupe:
I think the new version of react has problems with our use of reduxState. I don't know anything about redux and hardly about react, so trying to fix this has become a real trial-and-error process.


brdunfield:
Have you followed this?
https://github.com/reduxjs/react-redux/releases

There's some breaking changes in the update of react-redux to 6.0

smcmurtry:
Yikes it looks like store will no longer be passed as a prop to connected components in the new redux version. It's moving to the page context or something? That will break the way we are mocking redux in the tests. I've made upgrading redux its own issue: #1610

SupeDeDupe:
Yeah, the problem is that I don't know what store is actually for/doing in the first place beyond that it's redux and when I'm making a component, if it's not there, things break--which turns out to be a frustrating and problematic perspective to work from in trying to reformat the code to accommodate for the new react-redux upgrade.

smcmurtry (approved):
Looks good! 🦑

PR #1608, opened 2018-12-13
Fix ios zoom in on select

### LINKED ISSUE #1500: Share link copy button causes mobile browsers to magnify
See #1490 
Present on iOS

brdunfield:
Resolves #1500 

See: https://blog.osmosys.asia/2017/01/05/prevent-ios-from-zooming-in-on-input-fields/

Font size in a selected input needs to be 1em or higher to prevent the zoom in.


smcmurtry (approved):
Nice fix! 🦑

PR #1609, opened 2018-12-13
Display next steps from airtable

### LINKED ISSUE #1562: Display content from AirTable in the what's next section
Initially make not dynamic, just display all bullets

smcmurtry:
Closes #1562

### DIFF
--- a/__tests__/components/next_steps_test.js
+++ b/__tests__/components/next_steps_test.js
@@ -2,51 +2,47 @@ import React from "react";
 import { NextSteps } from "../../components/next_steps";
 import { mount } from "enzyme";
 import translate from "../fixtures/translate";
+import nextStepsFixture from "../fixtures/nextSteps";
 
 const { axe, toHaveNoViolations } = require("jest-axe");
 expect.extend(toHaveNoViolations);
 
 describe("NextSteps", () => {
   let props;
-  let reduxState;
   beforeEach(() => {
     props = {
       t: translate,
-      mapUrl: "/map"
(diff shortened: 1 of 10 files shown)


__tests__/components/next_steps_test.js
```diff
@@ -2,59 +2,55 @@ import React from "react";
...
+      nextSteps: nextStepsFixture
     };
-
-    reduxState = {};
-    props.reduxState = reduxState;
   });
 
   it("passes axe tests", async () => {
-    let html = mount(<NextSteps {...props} {...reduxState} />).html();
+    let html = mount(<NextSteps {...props} />).html();
```
smcmurtry, line 21: we don't actually need to pass in reduxState here since we aren't mocking it b/c there are not child components that use redux

utils/airtable_es2015.js
```diff
@@ -107,12 +107,6 @@ var hydrateFromAirtable = (exports.hydrateFromAirtable = async function hydrateF
     dataStore.multipleChoiceOptions,
     "variable_name"
   );
-  replaceId(
```
smcmurtry, line 110: This is unrelated, but I removed it because we aren't using this column any more

brdunfield (approved):
LGTM 👍 🦑

smcmurtry:
[Users are given options for what the next step is in their current stage](https://trello.com/c/rHNRWLYF/2-users-are-given-options-for-what-the-next-step-is-in-their-current-stage)

PR #1617, opened 2018-12-17
[WIP] Finish emotion Upgrade

brdunfield:
I've added the new https://emotion.sh/docs/@emotion/babel-preset-css-prop which should automatically do the jsx include / pragma on our components, but it doesn't seem to be working. Emotion is still working using className, but when I switch to the css prop I get an error that the include / pragma are missing..


brdunfield:
I've ripped out the new plugin and tried just straight updating everything with the jsx import and pragma, but it now it appears none of the composite classNames (that used to be composited with cx) are working.

brdunfield:
Closing because this is shaping up to be too much work - we will have the work archived and can revisit once emotion 10 has had more time / docs / examples to make this easier.

PR #1622, opened 2018-12-18
Changed font size to be the same as the save for later button

### LINKED ISSUE #1621: Make see more font same size as save for later, uncapitalize more/less
screen shot 2018-12-18 at 12 11 08 pm

smcmurtry:
Closes #1621


brdunfield (approved):
🦑

Issue #1626, opened 2018-12-18
IE - click "next steps" doesn't zoom you down the page 


courtneycl:
On IE Version 11.1446.15063.0. 

Clicking 'Next steps' doesn't scroll your browser down to where the next steps, like it does on Chrome. Expected it would.

brdunfield:
Related: #1630

brdunfield:
Closed via #1631

PR #1631, opened 2018-12-18
Refactor how scrolling to next-steps works - change to page anchor

### LINKED ISSUE #1630: Accessibility - Keyboard: Navigating to next steps via the next step button does not move the context to next steps
If you press tab again after navigating to next steps, the focus is still at the top of the page. Page scrolls back up.

brdunfield:
Closes #1630 and #1626

### DIFF
--- a/components/next_steps.js
+++ b/components/next_steps.js
@@ -81,7 +81,11 @@ export class NextSteps extends Component {
 
     return (
       <div className={outerDiv}>
-        <a href="next-steps" id="next-steps" />
+        <a
+          href="next-steps"
+          id="next-steps"
+          aria-label={t("nextSteps.whats_next")}
+        />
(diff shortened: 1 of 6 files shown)


components/next_steps.js
```diff
@@ -81,6 +81,11 @@ export class NextSteps extends Component {
 
     return (
       <div className={outerDiv}>
+        <a
```
smcmurtry: did you try adding the "next-steps" id to the header for the section? if so did it work? I think that would be preferable to a hidden element.

smcmurtry (approved):
Good work! Just a minor comment.

components/next_steps.js
brdunfield: It does work, I'll make that change before merging.

PR #1642, opened 2018-12-20
Fixed print page bug

### LINKED ISSUE #1527: Print page fonts
The print page first pops up in one font, and then changes font if the user clicks the cancel button on the print popup window. 

screen shot 2018-12-04 at 8 45 50 am
screen shot 2018-12-04 at 8 45 59 am

SupeDeDupe:
Closes #1527

### DIFF
--- a/components/header_link.js
+++ b/components/header_link.js
@@ -44,7 +44,6 @@ const grey = css`
 class HeaderLink extends Component {
   render() {
     const {
-      target,
       arrow,
       className,
       children,
@@ -52,7 +51,7 @@ class HeaderLink extends Component {
       size,
       altStyle,
       onClick,
-      otherProps
+      ...otherProps
     } = this.props;
 
     return (
@@ -65,7 +64,6 @@ class HeaderLink extends Component {
               ? cx(style, grey, className)
               : cx(style, className)
           }
-          target={target}
           href={href}
           onClick={onClick}
           {...otherProps}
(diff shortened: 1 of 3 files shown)


brdunfield:
This looks like it just loads with the default font, vs Merriweather?

SupeDeDupe:
Yep, that is correct. The main bug was that it dynamically changed fonts at some point between opening the page and closing the print window, if the print page was not opened in a new tab. Now that it always opens in a new tab, the user never sees it changing fonts in that way.

components/header_link.js
```diff
@@ -64,6 +65,7 @@ class HeaderLink extends Component {
               ? cx(style, grey, className)
               : cx(style, className)
           }
+          target={target}
```
smcmurtry: I think it would be better to leave `target` out of this file - the prop will be passed through automatically with `{...otherProps}`. Try reverting your changes to this component - the behaviour should be the same.

SupeDeDupe: If I remove changes from this file, I don't see how the print page will open in a new tab...

smcmurtry: Give it a try and see! Because of line 71, any prop passed into `HeaderLink` gets passed automatically to the `a` element.

SupeDeDupe: I tried removing line 68 along with other lines I had added, but this causes it to open in the same tab, not a new one.

smcmurtry: Okay it should be working now - the ... on line 54 was missing

smcmurtry (approved):
🦑

Issue #1478, opened 2018-11-27
When switching to a new page, voiceover doesn't announce the new page title or header


smcmurtry:
This happens both with next/link and with `setSection` `onClick` events in the guided experience.

smcmurtry:
![](https://github.trello.services/images/mini-trello-icon.png) [Accessibility: Screenreader Navigation](https://trello.com/c/knKtuM4T/100-accessibility-screenreader-navigation)

brdunfield:
@smcmurtry was this closed by #1649?

smcmurtry:
@brdunfield no that PR just solved the issue for /benefits-directory. The issue still exists for all the other pages.

smcmurtry:
Now that the GE has been split into pages this should be easier to solve. I've noticed that clicking the back link (not browser back) has the correct behaviour where the page title is announced and focus goes to the top of the page.

smcmurtry:
I've now tried putting autofocusing headers into the GE, but this doesn't seem to affect the voiceover focus ([which is different from the keyboard focus](https://webaim.org/discussion/mail_thread?thread=8526) btw). This is weird because I based the autofocusing off of what IRCC did and theirs does seem to affect the voiceover focus: https://vancouver.rescheduler-dev.cds-snc.ca/register

JuliannaR:
@smcmurtry if your having troubles, something could be a bug with voiceover, can you try validating with NVDA? If not, send it my way and I can see if I can help.

smcmurtry:
@JuliannaR If you could try to reproduce this issue with another screen reader that would be a huge help! I only have access to voiceover currently.

PR #1649, opened 2018-12-21
Focused header

smcmurtry:
Working on #1478. It will be complete once this component gets added to more pages.

I tried but couldn't successfully test the focus state.


brdunfield (approved):
LGTM! 🦑

PR #1653, opened 2019-01-03
export github data

sastels:
add commented out code to print the github data (to use in a presentation or whatnot). It didn't take me long to figure it out, but I thought it's probably worth it to add it to the code so others (including future Steve) don't have to figure it out again.


smcmurtry (approved):
Seems reasonable 🦑

PR #1665, opened 2019-01-07
Fix search losing focus

### LINKED ISSUE #1656: search bar loses focus while typing

smcmurtry:
Closes #1656 
 
I refactored `BenefitsPane` and split out 2 new components. This was to simplify the render logic to the point where I could use the same `Header` for both the case where `searchString === ""` and `searchString !== ""`. Using the same `Header` stops the component from being remounted and the autofocus triggering.


brdunfield (approved):
LGTM! 🦑

PR #1666, opened 2019-01-07
Improve skip to main content behaviour

### LINKED ISSUE #1636: Skip to main content button appears after the VAC header when navigating with VoiceOver

brdunfield:
Closes #1636 

I feel like the styling around this can still be improved, but this fixes the issue of the skip link not being the first thing announced by VoiceOver.

### DIFF
--- a/components/federal_banner.js
+++ b/components/federal_banner.js
@@ -67,11 +67,21 @@ const container = css`
   `)};
 `;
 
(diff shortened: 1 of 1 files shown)


smcmurtry (approved):
Looks good! Seems like there are a lot of different ways to do this but this way seems to work. 🦑

PR #1667, opened 2019-01-07
Fix for see more bug

### LINKED ISSUE #1664: See more button stops working if there are rows in benefitExamples with no linked benefit

smcmurtry:
Closes #1664 

Added a check that the linked_benefits key exists on the example content returned from airtable

I tested this by adding in a row to benefitExamples with missing linked benefits. I deleted the row so that see more would not be broken in master.


brdunfield (approved):
👍

PR #1668, opened 2019-01-08
Added a second GA tracker id so both CDS and VAC get analytics

### LINKED ISSUE #1661: Add a 2nd google analytics tracking code so both CDS and VAC can capture data
The tracking code should be passed in a 2nd environment variable. I already added it to circle ci: GA_UA_CDS.

This article might help: https://support.google.com/analytics/answer/1032400?hl=en

smcmurtry:
Closes #1661 

Tested locally by hardcoding GA_UA_CDS and checking that I was visible in google analytics dashboard.


brdunfield (approved):
🎉 🦑

PR #1670, opened 2019-01-08
Demoted edit selections from h2 to div

### LINKED ISSUE #1405: Headings are out of order in DOM on benefits-directory page
Using keyboard navigation to go through the benefits-directory page follows DOM order. Currently you get to an h2 before the h1, which is a problem. I'm not sure how to fix this because with side by side grid containers, the thing on the left is first in the DOM order.

SupeDeDupe:
I don't think Edit Selections needs to be H2, especially if it's going to be removed soon.

Closes #1405

### DIFF
--- a/components/needs_selector.js
+++ b/components/needs_selector.js
@@ -37,9 +37,7 @@ export class NeedsSelector extends Component {
     if (showQuestion("needs", undefined, this.props.reduxState)) {
       return (
         <div className={topBorder}>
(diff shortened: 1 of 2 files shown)


brdunfield:
When the edit selections menu is expanded, "Filter by category" is an H3 that comes before the main H1.

smcmurtry (approved):
Looks good! 🦑

PR #1672, opened 2019-01-08
Add group names to radio buttons for accessibility / navigation

### LINKED ISSUE #1640: Radio button groups get combined in the benefits directory filters
When using VoiceOver, it perceives the separate button groups as a single group. Tab functionality also behaves as if they are a single group. This makes it difficult to navigate with Tab and understand with VoiceOver
screen shot 2018-12-20 at 11 47 53 am

brdunfield:
Closes #1640


smcmurtry (approved):
Great work Bethany!! that is such a huge improvement 🦑

PR #1676, opened 2019-01-09
Fixed language change bug in benefit cards

### LINKED ISSUE #1671: Cards need to be refreshed to show language change
If you have already pre-selected filters and then select toggle language English -> Français the cards still display the original language until you hit refresh. This happens when you switch back to English as well. 
All of the other page text translates properly when you click the language button. 
Note: This does not happen if you change language before selecting filters.

To reproduce filter on benefits dir page
Select:
   -- Benefits for family members
   -- Royal Canadian Mounted Police
   -- Release
Click on Français

The cards are still in English until you hit refresh

SupeDeDupe:
Closes #1671

### DIFF
--- a/__tests__/components/benefit_card_test.js
+++ b/__tests__/components/benefit_card_test.js
@@ -38,6 +38,7 @@ describe("BenefitCard", () => {
     props = {
       t: () => "en",
       benefit: benefitsFixture[0],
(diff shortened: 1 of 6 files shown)


smcmurtry (approved):
Nice work! 🦑

Issue #1660, opened 2019-01-07
Do an eligibility filter on the "also eligible for" benefits in the card expansion


smcmurtry:
We could do this both for veteran's and family members' benefits. We may need to send in different profile selections to the filter if we are a vet getting the family benefits, or a family member getting the vet benefits. This may require a refactor of how the filter selectors work, or maybe making a new function/selector that reuses some of the existing filtering logic.

smcmurtry:
We added in and then deleted a test version of the new Disability Benefits card. Here is the data from airtable:

```Disability Benefits	rectQ6WL0wvRgn3de	Compensation for your service-related injury or illness.	(fra) Disability Benefits	Une indemnité pour les vétérans ayant des problèmes de santé causés ou aggravés par leur service.	http://www.veterans.gc.ca/eng/services/after-injury/disability-benefits/disability-award	http://www.veterans.gc.ca/fra/services/after-injury/prestations-invalidite/disability-award			Career Impact Allowance, Caregiver Recognition Benefit, Clothing Allowance, Funeral and Burial Assistance, Treatment Benefits, Veterans Independence Program, Attendance Allowance, Survivor's Pension, Exceptional Incapacity Allowance	Independent	Income assistance, Medical expenses	"veteran, servingMember, RCMP, hasServiceHealthIssue", "veteran, WSV (WWII or Korea), hasServiceHealthIssue", "veteran, servingMember, CAF, hasServiceHealthIssue"	high		```

SupeDeDupe:
![](https://github.trello.services/images/mini-trello-icon.png) [Create Disability Benefits card by merging Disability Award and Disability Pension](https://trello.com/c/jVT3DHR7/138-create-disability-benefits-card-by-merging-disability-award-and-disability-pension)

PR #1677, opened 2019-01-09
Refactored benefit expansion component

smcmurtry:
Working on #1660
I did this to make it easier to implement filtering of "also eligible" benefits
There should be no changes in behaviour of the app


SupeDeDupe (approved):
I think moving those parts into components was a good idea.
Great work! 👯 🦑

PR #1679, opened 2019-01-09
Another attempt to get 2 GA trackers set up 

### LINKED ISSUE #1678: Get google analytics working for real this time
Although it was working when we set the `GA_UA_CDS` env variable in heroku, it doesn't work when we set both `GA_UA` and `GA_UA_CDS`.

smcmurtry:
Closes #1678 

For some reason the solution is to add a name to one of the trackers but not the other?! Oh well it seems to work.

### DIFF
--- a/utils/analytics.js
+++ b/utils/analytics.js
@@ -2,8 +2,14 @@ import ReactGA from "react-ga";
 
 export const initGA = () => {
   ReactGA.initialize([
-    { trackingId: process.env.GA_UA },
-    { trackingId: process.env.GA_UA_CDS }
+    {
+      trackingId: process.env.GA_UA,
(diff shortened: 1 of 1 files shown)


utils/analytics.js
```diff
@@ -1,10 +1,18 @@
...
+    [
+      {
+        trackingId: process.env.GA_UA,
+        gaOptions: { name: "tracker_1" }
+      },
+      {
+        trackingId: process.env.GA_UA_CDS
+      }
+    ]
+    // , { debug: true }
```
smcmurtry, line 14: leaving this in since it will be useful if we ever need to modify this file again

SupeDeDupe (approved):
Good job 🦑 👍

PR #1681, opened 2019-01-09
Filter "see more"/"also eligible for" benefits by eligibility

### LINKED ISSUE #1660: Do an eligibility filter on the "also eligible for" benefits in the card expansion

smcmurtry:
Closes #1660 

I've tested this somewhat thoroughly - the only place I can see a difference in the "see more" benefits here vs. master is for Disability Pension with Veteran + RCMP selected, the family is eligible for 1 less benefit. But I think that is an improvement since it better reflects the eligibility requirements. Otherwise this should have the same benefits in "see more" as master.


brdunfield (approved):
Awesome! 🦑

PR #1687, opened 2019-01-11
added option to anonymize IPs collected by google analytics

smcmurtry:
Closes #1687 

I'm not sure if it's possible to verify that this is working, because IPs are not visible in the Google Analytics dashboard. But I did verify that location data is unaffected by this change.

### DIFF
--- a/utils/analytics.js
+++ b/utils/analytics.js
@@ -13,6 +13,7 @@ export const initGA = () => {
     ]
     // , { debug: true }
   );
+  ReactGA.set({ anonymizeIp: true });
 };
 
 export const logPageView = () => {


SupeDeDupe (approved):
Can test it by verifying aip=1 in the query params. Can't see it in this heroku deployment though (maybe the GA setup is attached to the live app domain?)

#1687:
screen shot 2019-01-14 at 10 03 55 am

live app:
screen shot 2019-01-14 at 10 04 03 am

https://www.optimizesmart.com/how-to-turn-on-ip-anonymization-in-google-analytics-and-google-tag-manager/

PR #1694, opened 2019-01-11
Refactor eligibility paths

### LINKED ISSUE #1680: Read eligibility information from new AirTable sheet
Instead of the legacy eligibilityPaths

brdunfield:
Closes #1680 

This changes over the eligibility matching in the benefits selector for the benefit cards to the new benefitEligibility format / table from airtable, instead of the old eligibilityPaths.

Old eligibility match code has not yet been removed as other content such as next steps is still using the old format.

### DIFF
--- a/__tests__/components/BB_test.js
+++ b/__tests__/components/BB_test.js
@@ -7,6 +7,7 @@ import configureStore from "redux-mock-store";
 import { BB } from "../../components/BB";
 import benefitsFixture from "../fixtures/benefits";
 import eligibilityPathsFixture from "../fixtures/eligibilityPaths";
+import benefitEligibilityFixture from "../fixtures/benefitEligibility";
 import needsFixture from "../fixtures/needs";
 import areaOfficesFixture from "../fixtures/area_offices";
 import questionsFixture from "../fixtures/questions";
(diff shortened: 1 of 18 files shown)


SupeDeDupe:
After talking a bit with Allison about next steps admin editing options, it seems like maybe we should link benefits from benefitEligibility in nextSteps in Airtable, and have additional fields linking to multipleChoiceOptions so that broader categories of benefits can be associated with certain next steps.

smcmurtry:
I'm getting a different number of results for this PR vs master:

https://vac-poc-pr-base-pr-1694.herokuapp.com/benefits-directory?lng=en&patronType=veteran&serviceType=CAF

vs. 

https://vac-benefits-finder.cds-snc.ca/benefits-directory?lng=en&patronType=veteran&serviceType=CAF

SupeDeDupe:
@smcmurtry That's because of the new benefits Allison added this week to Airtable, which haven't been configured into benefitEligibility yet. 

The title also isn't showing up properly, but I believe it's for the same reason.

screen shot 2019-01-16 at 9 12 24 am

SupeDeDupe:
The title also isn't showing up properly, but I believe that's also because of the new benefits that haven't yet been added to benefitEligibility. 

screen shot 2019-01-16 at 9 12 24 am

SupeDeDupe (approved):
Update: The Airtable sheet has been updated and those problems have been fixed.
Good job 🥇🦑

brdunfield:
[Change eligibility path structure to be more like Allison's spreadsheet](https://trello.com/c/KXov9hiW/148-change-eligibility-path-structure-to-be-more-like-allisons-spreadsheet)

PR #1697, opened 2019-01-11
Family benefit fix

### LINKED ISSUE #1695: Family benefits that require deceased should not show in see more list
This includes many of the benefits family members are (supposedly) eligible for in the disability benefits card. This may be tricky to fix because the stillServing/released/deceased question is only asked to family members, and the card only shows if you are not a family member. What to do?

smcmurtry:
Closes #1695


SupeDeDupe (approved):
Good job! 🦑

PR #1703, opened 2019-01-14
Made skip to main content button its own component, fixed language button bug

### LINKED ISSUE #1702: Language change button has a reduced clickable area because it's being overlapped by another element

smcmurtry:
Closes #1702 

The language button is no longer obscured. So that the skip link will be read out by voiceover I made the div containing the federal+beta banners a header element (I got the idea from this page: https://www.w3.org/TR/wai-aria-practices/examples/landmarks/HTML5.html#). Voiceover focuses on the header initially, and then tabbing gets you to the skip button.


brdunfield (approved):
LGTM! 🦑

PR #1706, opened 2019-01-15
Moved refresh cache button from header on /data-validation

### LINKED ISSUE #1696: Move Refresh Cache button from the top bar to the main content area of data-validation
The button is not very usable right now because of another element in front of it. Might as well just move it since it only needs to be on the data-validation page anyway. And we could change it to a govuk based button rather than the material-ui button while we're at it.

smcmurtry:
Closes #1696


SupeDeDupe (approved):
Nice 👌 🦑

PR #1707, opened 2019-01-15
Added an anchor element to learn more button

### LINKED ISSUE #1705: Use an anchor element for Learn More buttons
From John O'Brien:

"On the list of all benefits pages, the green buttons that link out from a particular benefit exhibit.. non-traditional link based behaviours.

1: There is no mouse-over view of the domain in the browser status bar at bottom of page.
2: Right clicking on button does not present the typical (open in..) experience that one expects with an <a href=>

These combined lead to what I would consider is not-recommended experience in that the user
has no way to determine the safety of clicking on the button. "

smcmurtry:
Closes #1705 

The benefit of this is that a user can now see the href when mousing over the button. And middle click/command+click/ctrl+click now works.

### DIFF
--- a/__tests__/components/learn_more_button_test.js
+++ b/__tests__/components/learn_more_button_test.js
@@ -7,9 +7,6 @@ expect.extend(toHaveNoViolations);
 
 describe("LearnMoreButton", () => {
   let props;
-  const mocked_fn = jest.fn();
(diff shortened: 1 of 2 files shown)


brdunfield (approved):
🦑 nice!

PR #1709, opened 2019-01-15
See more content is now searchable

### LINKED ISSUE #1657: See more text not returned in search

smcmurtry:
Closes #1657 

You should now get results if your search includes text that is in the see more example list. The matching text is highlighted (although you still have to expand the card to see it).

I didn't update the french search index fixture because when I did I got an error trying to load the file. Not sure why, maybe some kind of encoding issue? Anyway, I'll make a separate issue for that.


SupeDeDupe:
Should we be expanding cards with highlighted words in the see more, so the user can see how the result matched their search?

smcmurtry:
I think there are pros and cons to doing that. Probably a good question to involve design in. I think it would be a separate PR in any case.

SupeDeDupe (approved):
👍🦑

PR #1710, opened 2019-01-16
See more display for "deceased" statusAndVitals

### LINKED ISSUE #1701: Do not show child benefits a service member may qualify for if they are deceased
screen shot 2019-01-14 at 10 39 10 am

SupeDeDupe:
Closes #1701

### DIFF
--- a/components/benefit_expansion.js
+++ b/components/benefit_expansion.js
@@ -49,7 +49,7 @@ export class BenefitExpansion extends Component {
   };
 
   render() {
-    const { t, benefit, benefits, store } = this.props;
+    const { t, benefit, benefits, store, reduxState } = this.props;
     const language = t("current-language-code");
     const benefitName =
       language === "en" ? benefit.vacNameEn : benefit.vacNameFr;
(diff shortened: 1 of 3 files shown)


brdunfield:
This looks good, but i'm not the biggest fan of the code duplication in the render function - the two branches in render are identical aside from the extra ChildBenefitList.

components/benefit_expansion.js
```diff
@@ -72,12 +72,28 @@ export class BenefitExpansion extends Component {
...
+    const profileFilters = JSON.parse(
+      JSON.stringify(getProfileFilters(reduxState, this.props))
+    );
 
     let otherBenefits = t("benefits_b.eligible_open_veteran", {
       x: benefitName
     });
 
-    return (
+    return profileFilters["statusAndVitals"] === "deceased" ? (
```
smcmurtry: You don't need to use the `profileFilters` object here. You can just use `reduxState.statusAndVitals`. The `profileFilters` object just exists so we can loop over all the profile questions easily, but all the same data is in redux.

smcmurtry: I agree with @brdunfield that there's a simpler way to do this. We just need to prevent the service member list from rendering if they are deceased. If you look at the `ChildBenefitList` component, it doesn't render if it is passed an empty list as the benefits prop. So you could set the `vetServBenefits = []` if `reduxState.statusAndVitals === "deceased"`.

SupeDeDupe:
I didn't like the code duplication either

smcmurtry (approved):
Looks good! 🦑

Issue #1693, opened 2019-01-11
Convert remaining eligibilityPaths code to new benefitEligibility format


brdunfield:

SupeDeDupe:
![](https://github.trello.services/images/mini-trello-icon.png) [Change eligibility path structure to be more like Allison's spreadsheet](https://trello.com/c/KXov9hiW/148-change-eligibility-path-structure-to-be-more-like-allisons-spreadsheet)

smcmurtry:
Closed by #1722

PR #1713, opened 2019-01-16
Convert next steps to stop using Eligibility Paths

### LINKED ISSUE #1692: Convert other AirTable tables to new benefitEligibility format

brdunfield:
Closes #1692 and #1693

### DIFF
--- a/selectors/benefits.js
+++ b/selectors/benefits.js
@@ -4,7 +4,6 @@ import { createSelector } from "reselect";
 const getBenefits = state => state.benefits;
 const getQuestions = state => state.questions;
 const getCurrentLanguage = (state, props) => props.t("current-language-code");
-const getEligibilityPaths = state => state.eligibilityPaths;
 const getBenefitEligibility = state => state.benefitEligibility;
 const getMultipleChoiceOptions = state => state.multipleChoiceOptions;
 const getEnIdx = state => state.enIdx;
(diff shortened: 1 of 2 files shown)


selectors/benefits.js
```diff
@@ -183,40 +181,35 @@ export const getFilteredBenefits = createSelector(
...
       if (profileFilters[criteria] !== "") hasSelections = true;
     });
 
-    if (hasSelections)
+    if (hasSelections) {
+      // only check eligible nextSteps if we know we have selections
+      let eligibleNextStepIds = [];
+      nextSteps.forEach(ns => {
+        if (
+          benefitEligibilityMatch(ns, profileFilters, multipleChoiceOptions)
```
smcmurtry, line 197: maybe we should rename this function to be `eligibilityMatch` if it's going to be used for next steps too.

smcmurtry (approved):
Looks good, one minor comment. I think we should eventually make a separate nextStepsEligibility sheet so we can have multiple rules for a single next step.

selectors/benefits.js
brdunfield: I think that this could be done as part of an upcoming PR to remove eligibilityPaths from the code.

brdunfield:
[Change eligibility path structure to be more like Allison's spreadsheet](https://trello.com/c/KXov9hiW/148-change-eligibility-path-structure-to-be-more-like-allisons-spreadsheet)

PR #1715, opened 2019-01-16
Adding missing spread operator

### LINKED ISSUE #1655: Add missing spread operator to components that are trying to use it incorrectly (the ... operator)

smcmurtry:
Closes #1655 

The spread operator is used in these components is to pass through any other props that are sent, without having to hard code them in. Now that should actually work.


SupeDeDupe:
I think there's another one in search_box.js on line 118

smcmurtry:
Thanks @SupeDeDupe! It should be fixed now. I also removed a component we weren't using any more.

SupeDeDupe (approved):
Awesome! 🦑

PR #1719, opened 2019-01-17
Relevant question eligibility

brdunfield:
I think this is the last code logic that needed to be ported over - towards #1693

### DIFF
--- a/__tests__/components/benefit_list_test.js
+++ b/__tests__/components/benefit_list_test.js
@@ -6,6 +6,7 @@ expect.extend(toHaveNoViolations);
 
 import benefitsFixture from "../fixtures/benefits";
 import eligibilityPathsFixture from "../fixtures/eligibilityPaths";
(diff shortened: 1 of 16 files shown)


__tests__/components/benefit_list_test.js
```diff
@@ -6,6 +6,7 @@ expect.extend(toHaveNoViolations);
 
 import benefitsFixture from "../fixtures/benefits";
 import eligibilityPathsFixture from "../fixtures/eligibilityPaths";
```
SupeDeDupe, line 8: Should anything containing "eligibilityPaths" still be in the code at all, then??

SupeDeDupe: I noticed it's in quite a few more places after this one, and I'm not sure if that was on purpose or not.

SupeDeDupe (approved):
Good work! 👍

PR #1720, opened 2019-01-18
Fixed IE invisible bullet point bug

### LINKED ISSUE #1627: IE - missing second column bullets for see more content
What Michelle saw with her earlier version of Chrome! 

IE v11.1446.15063.0 

image

smcmurtry:
Closes #1627 

Tested with browserstack. I added a comment so this doesn't accidentally get removed in the future since it seems like it's not doing anything.


SupeDeDupe (approved):
👍

PR #1738, opened 2019-01-22
Auto fill airtable

### LINKED ISSUE #1732: Auto fill empty airtable values
There have now been a few cases where the app was brought down by missing values in airtable. The issue is that if a cell is empty, the airtable api sends the row object back without a key for the empty column. So if we are checking for the column and assuming it's there, this can cause the app to crash.

To fix this we should auto fill empty values on the server, so that a key exists for every column. For example, if col2 has no value, we currently have:  `{col1: "asdf", col3: "asdf"}`. Instead we should make it: `{col1: "asdf", col2: "", col3: "asdf"}`.

smcmurtry:
Closes #1732 

Tested by adding a row to the benefitEligibility sheet. Again it broke master (500 error on summary page with no selections) but not this branch.

### DIFF
--- a/utils/airtable_es2015.js
+++ b/utils/airtable_es2015.js
@@ -60,6 +60,29 @@ var fetchTableFromAirtable = async function fetchTableFromAirtable(table) {
   });
 };
 
+var fillEmptyValues = table => {
+  var array = table.map(x => Object.keys(x).length);
+  var number_of_fields = Math.max(...array);
+  var completed_row = table.filter(
+    x => Object.keys(x).length === number_of_fields
+  )[0];
+  var all_columns = Object.keys(completed_row);
+
+  table.forEach(x => {
+    var x_columns = Object.keys(x);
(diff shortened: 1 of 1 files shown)


utils/airtable_es2015.js
```diff
@@ -72,20 +95,7 @@ var hydrateFromAirtable = (exports.hydrateFromAirtable = async function hydrateF
   await Promise.all(promises);
   dataStore["errors"] = [];
   airtableConstants.tableNames.forEach(function(tableName) {
-    var array = dataStore[tableName].map(x => Object.keys(x).length);
```
smcmurtry, line 75: I removed this because it was kind of a dumb algorithm that sometimes resulted in unexpected consequences. What it did: if less than half of the columns in a row are filled, that row is rejected and an error is sent to airtable.

SupeDeDupe (approved):
Awesomee 💃

PR #1747, opened 2019-01-24
Make see more content sentence dynamically read from airtable

### LINKED ISSUE #1737: Make the 1st sentence of see more content configurable in airtable
Currently this sentence is hardcoded as "Some examples of what this program may provide:". We want to create some more flexibility so some benefits can have different lists. An example is as disability benefits: here we want to list examples of health issues that would make you eligible.

brdunfield:
Closes #1737 

Added two new columns to the benefits table in AirTable: seeMoreSentenceEn and seeMoreSentenceFr, where this sentence can be customized.

### DIFF
--- a/components/example_bullets.js
+++ b/components/example_bullets.js
@@ -49,7 +49,7 @@ export class ExampleBullets extends React.Component {
   };
 
   render() {
-    const { t } = this.props;
+    const { t, benefit } = this.props;
     const bullets = this.getExampleBullets();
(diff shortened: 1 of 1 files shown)


smcmurtry (approved):
Nice work! Only suggestion would be to add this to the benefits fixture and add a test for it.

brdunfield:
I will do that as a separate PR

Issue #1733, opened 2019-01-22
Back button in GE doesn't behave as expected


smcmurtry:
Clicking back in GE doesn't take you to the previous screen because all the initial questions are the same page.

We can fix this by either:
a) splitting the questions into individual pages
b) pushing to the browser history when we switch section in GE

smcmurtry:
Closed by #1748

PR #1748, opened 2019-01-24
GE refactor

smcmurtry:
Working on #1733, this should also make #1478 easier

This is my attempt at a massive refactor of the guided experience with the goal of making it simpler by turning each question into its own page. What do you think - is this simpler?


brdunfield:
I know this is still [WIP], but I notice that selected data gets cleared when you go back - is that desirable behaviour? Like if you select "veteran" and proceed forward, when you navigate back using the in-app back button that selection persists, but if you use the browser back it gets cleared.

smcmurtry:
@brdunfield It looks like that is happening because I'm not immediately putting the user's selection into the url query parameters. It gets put into the url when you visit the next page. So when you click browser back, it goes back to a url that doesn't include the query param, and then that clears it from redux. 

I tried changing the behaviour to be more like it was before, but it started introduced a lot of complexity to the code. So my preference would be to leave it like this until we get feedback from user testing that it's an issue. If someone hits browser back they probably wanted to re-answer the question anyway.

brdunfield:
Tests are passing but getting a few failed propTypes for fontSize on AnchorLink in the tests.

brdunfield (approved):
👍 🦑 Nice work!

PR #1750, opened 2019-01-24
Updating sidebar for mobile

### LINKED ISSUE #1726: Sticky header for mobile in benefits directory
https://app.zeplin.io/project/5acbd659dc553ec84805afd4/screen/5c3f8e799f82f718aea01b56
screen shot 2019-01-21 at 3 12 43 pm

SupeDeDupe:
Closes #1726 and #1754

### DIFF
--- a/components/BB.js
+++ b/components/BB.js
@@ -42,11 +42,8 @@ const sidebar = css`
   position: -webkit-sticky;
   position: sticky;
   top: 0;
-  @media only screen and (max-width: ${globalTheme.max.xs}) {
-    position: -webkit-sticky;
-    position: sticky;
-    top: 0;
-  }
+  background-color: ${globalTheme.colour.white};
+  z-index: 1;
 `;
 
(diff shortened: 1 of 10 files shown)


SupeDeDupe:
Need to set the offset in the scroll to only come into effect on mobile, then I think that'll be the end of the PR.

components/BB.js
```diff
@@ -77,10 +101,21 @@ export class BB extends Component {
   }
   scrollToNextSteps() {
     window.location = "#next-steps";
+    const maxMobile = parseFloat(globalTheme.max.xs);
+    window.screen.width < maxMobile ? window.scrollBy(0, -90) : null;
```
smcmurtry, line 105: I tested with my phone and this offset is working great! Can you apply it to the favourites page too?

SupeDeDupe: Oh yes, I meant to not forget about that...

smcmurtry (approved):
Nice work! 🦑

PR #1760, opened 2019-01-25
Move focus for screen readers when "See More" button is clicked

### LINKED ISSUE #1652: Keyboard navigation after pressing "See More" can be improved
When using keyboard navigation, after selecting a "see more" button, the focus is still on the see more button, which is below the content that has just been expanded.

This can be improved somehow.

brdunfield:
Closes #1652

### DIFF
--- a/components/benefit_expansion.js
+++ b/components/benefit_expansion.js
@@ -1,6 +1,6 @@
 import React, { Component } from "react";
 import PropTypes from "prop-types";
-import { css } from "emotion";
+import { cx, css } from "emotion";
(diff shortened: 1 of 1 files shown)


smcmurtry:
I think the behaviour is great but the blue focus box that shows up every time you click see more is weird. Could you disable the blue focus style for the div with some css? 
```
const noFocus = css`
  :focus {
      outline: none;
  }`;
```

smcmurtry (approved):
Nice! 🦑

PR #1762, opened 2019-01-28
Fix for column wrapping in IE

### LINKED ISSUE #1731: IE - Long bullets overflow into second column instead of wrapping
Group Health Insurance only has one bullet that spans two lines. Chrome wraps the bullet. IE extends it into the second column. 

Ideally, if there is only one column worth of bullets, there would be no columns and the text could extend across the card with standard spacing. A quicker fix is to make this wrap in all browsers. 

Chrome: 

image

IE: 

image

brdunfield:
Closes #1731

### DIFF
--- a/components/child_benefit_list.js
+++ b/components/child_benefit_list.js
@@ -26,6 +26,7 @@ const heading = css`
 `;
 const listStyle = css`
   padding-left: 20px;
+  text-indent: 0.2em;
 `;
 
 const logExit = url => {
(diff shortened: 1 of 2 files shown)


smcmurtry:
Looks good! My only suggestion is that you make the indentation on the see more bullets and the benefits list the same:
screen shot 2019-01-28 at 5 21 12 pm

smcmurtry (approved):
Nice work! 🦑

PR #1774, opened 2019-01-29
Added helper text back to service health issue page

### LINKED ISSUE #1772: Descriptive text missing from health issue page
This probably happened with the GE refactoring

smcmurtry:
Closes #1772

### DIFF
--- a/__tests__/components/guided_experience_test.js
+++ b/__tests__/components/guided_experience_test.js
@@ -77,17 +77,12 @@ describe("GuidedExperience", () => {
     ).toContain("next");
   });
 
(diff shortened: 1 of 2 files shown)


brdunfield:
This looks good, but I'm wondering if the display around the tooltip should be around whether question.tooltip_english / french exists and not whether it's the serviceHealthIssue page - just in case tooltips are added / removed to other questions down the line? Thoughts?

smcmurtry:
@brdunfield yes that's a good idea. I'll change it

SupeDeDupe (approved):
Looks good! 🦑

PR #1777, opened 2019-01-30
Fixed getSummaryUrl Selector

### LINKED ISSUE #1773: Editing selections after a search has been made gives the wrong number of results on summary page
https://vac-benefits-finder.cds-snc.ca/benefits-directory?lng=en&searchString=asd&patronType=veteran&serviceType=CAF

Then click edit selections and note that the number of results changes on the summary page depending on your search term.

Maybe we should clear the search term when someone clicks "edit selections". Is there any case when we wouldn't want to clear the search term when a link is followed?

SupeDeDupe:
Closes #1773

### DIFF
--- a/__tests__/components/BB_test.js
+++ b/__tests__/components/BB_test.js
@@ -124,4 +124,21 @@ describe("BB", () => {
       .simulate("click");
     expect(mounted_BB().instance().scrollToNextSteps).toBeCalled();
   });
+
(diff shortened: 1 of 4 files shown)


smcmurtry (commented):
Works great! My only suggestion is to add a test for this.

smcmurtry (approved):
Great work! 🦑

PR #1778, opened 2019-01-30
Don't render next steps if there are none

### LINKED ISSUE #1755: Next steps header should not be displayed if the list is empty
image

brdunfield:
Closes #1755 . Decided to implement it anyways since it was like a one line fix.

Currently there's no way to test this in the app, since there are 3 next steps in Airtable with no requirements.

I wrote a test for it in jest. If you want to test it in the app, you'll need to modify getBullets() to return an empty array.

### DIFF
--- a/__tests__/components/BB_test.js
+++ b/__tests__/components/BB_test.js
@@ -50,7 +50,8 @@ describe("BB", () => {
       filteredBenefits: [],
       setSelectedNeeds: () => true,
       favouriteBenefits: [],
-      url: { query: {} }
+      url: { query: {} },
+      filteredNextSteps: nextStepsFixture
     };
     _shallowBB = undefined;
     _mountedBB = undefined;
--- a/__tests__/components/favourites_test.js
+++ b/__tests__/components/favourites_test.js
@@ -48,7 +48,8 @@ describe("Favourites", () => {
       },
       url: { query: {} },
       homeUrl: "/",
-      nextStepsRef: React.createRef()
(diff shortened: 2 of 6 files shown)


SupeDeDupe:
I just tested it by temporarily editing the records in Airtable. It might be a concern that the next steps link in the sidebar is still there and doesn't redirect to anywhere.

brdunfield:
That's true. It probably shouldn't render in the sidebar if there are no next steps. If it's not too much work I'll update it. But the case of having no next steps probably shouldn't happen so I'm not sure how much work to put into it.

smcmurtry:
I don't think we need to worry about how the app looks if there are no next steps. If there are no next steps, the issue is that they are missing from airtable and should be added. Is there any legitimate case where the user wouldn't be eligible for any next steps? If so that is an edge case that should probably have its own design rather than just hiding things.

SupeDeDupe:
Yeah, at the time when I made the issue originally, I wasn't thinking that the empty list was a bug. Seems like we're getting into a bit of a rabbit hole, so I'd also agree that perhaps we don't need to worry about the case of the list being empty as it would be a content bug....

brdunfield:
So should we just stop this PR here, or is there additional functionality we want in it?

smcmurtry:
My preference is to close without merging.

smcmurtry:
whoops! didn't mean to close it myself :)

brdunfield:
Works for me

PR #1783, opened 2019-01-30
Remove stats page

### LINKED ISSUE #1781: Remove stats page because we aren't using it any more

smcmurtry:
Closes #1781


sastels (approved):
😭 
LGTM, I think you got everything. One thing, we should probably change data-validation page (another issue/PR) so we don't need react-moment. probably heavy to just display a date.

PR #1785, opened 2019-01-30
Remove deleted benefits on page load of either directory or favourites

### LINKED ISSUE #1779: Benefits that are removed from Airtable persist in cookies
If a user saves a benefit to their saved benefits list, and the benefit is later removed from Airtable, it will persist as a saved benefit in their cookies. The Saved List number indicator will count the benefit, but when the Favourites page is loaded the benefit will not display, so the indicator number will not match the number of benefits displayed on the favourites page.

To test, will most likely want to make an Airtable copy so that we aren't adding / removing benefits from master.

brdunfield:
Closes #1779

On page load of either the benefits directory or the favourites page, the app checks the current favourite benefits in state against the full list of benefits, prunes any that have been deleted, and replaces the cookie.

I'm not happy with this yet - on page load the first (incorrect) number still appears for a couple seconds while the app is loading before being replaced with the correct number. I think the check needs to happen further up the pipeline. Any thoughts? @smcmurtry


smcmurtry:
@brdunfield I made a commit that I think will work for pruning out outdated favourite benefits on the server. But I haven't tested it yet so no guarantees it will work :)

brdunfield:
Do you think it's an issue if the deleted benefits technically persist on the client cookies but are pruned by the server? The user facing app will work fine, it's just the data on the client will contain the extra ids.

brdunfield:
Setting cookies.get() in server.js did not work..

brdunfield:
I have re-added the functionality to favourites and BB but added a check to see if the cookie exists before it attempts a prune.

This allows tests to pass. I still need to write some tests to actually test the functionality.

utils/airtable_es2015.js
```diff
@@ -5,9 +5,9 @@ const Logger = require("./logger").default;
 exports.hydrateFromAirtable = exports.writeFeedback = undefined;
 
 var airtableConstants = require("./hardcoded_strings");
-var readKey = process.env.AIRTABLE_READ_KEY;
+var readKey = "keyj9WrZiZ6OL7nfp"; //process.env.AIRTABLE_READ_KEY;
 var writeKey = process.env.AIRTABLE_WRITE_KEY;
-var baseKey = process.env.AIRTABLE_BASE_KEY || "appoFDwVvNMRSaO6o";
+var baseKey = "appKREUCgIezNnqJF"; //process.env.AIRTABLE_BASE_KEY || "appoFDwVvNMRSaO6o";
```
smcmurtry: Don't forget to revert the changes to this file :)

smcmurtry (approved):
Great work! 🦑

PR #1787, opened 2019-01-31
[WIP] fixing focus in GE

smcmurtry:
Working on #1478

This is only a PR so I can test the behaviour in the real app since it seems to be different from the dev version.


smcmurtry:
Closing this since I didn't find a solution

PR #1791, opened 2019-02-01
remove react-moment

### LINKED ISSUE #1790: remove react-moment module
Now that the stats page is gone, we should probably refactor the one line of the data-validation page that uses react-moment so that we don't need to use that module.

sastels:
fixes #1790 . refactor the data-validation page to not use react-moment. Basically, time of last Airtable cache used to look like

image

now looks like

image


SupeDeDupe (approved):
Nice, but I'm going to miss the AM / PM timestamps 🦑

PR #1806, opened 2019-02-05
GE skip button bug

### LINKED ISSUE #1801: Clicking skip from 1st page in GE w/ veteran selected should go to needs
But instead it goes to service type.

smcmurtry:
Closes #1801 

The issue was that we were getting the next section using the redux state, which may be incorrect for the current question if skip is clicked (since clicking skip should clear the answer for the current question). Also one of our tests was wrong.

### DIFF
--- a/__tests__/components/guided_experience_test.js
+++ b/__tests__/components/guided_experience_test.js
@@ -137,8 +137,6 @@ describe("GuidedExperience", () => {
         .find("#skipLink")
         .first()
         .props().href
-    ).toEqual(
-      "/serviceHealthIssue?lng=en&patronType=veteran&selectedNeeds=1,2"
-    );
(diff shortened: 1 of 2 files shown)


__tests__/components/guided_experience_test.js
```diff
@@ -137,8 +137,6 @@ describe("GuidedExperience", () => {
         .find("#skipLink")
         .first()
         .props().href
-    ).toEqual(
-      "/serviceHealthIssue?lng=en&patronType=veteran&selectedNeeds=1,2"
```
SupeDeDupe, line 141: lol what even was this test

SupeDeDupe (approved):
Awesome! Good call on adding those functions 🦑🥇

PR #1811, opened 2019-02-06
Airtable download script

### LINKED ISSUE #1802: yarn download
write a node script to download airtable tables and put in flat json files

sastels:
resolves #1802 

Add a `yarn download` command that downloads the airtable data to `data/data.json`


smcmurtry (approved):
Tested locally and it seems to work fine. Thanks Steve!

PR #1815, opened 2019-02-06
Use static airtable data in production

### LINKED ISSUE #1803: use airtable flat files
once #1802 merges, change the app so that in prod mode it uses the flat files rather than caching from Airtable

sastels:
resolves #1803

Changes the default (production and PR review) app to
- use the static dump of airtable at `data/data.json`
- not serve the data validation page

I've created a staging app at https://vbd-staging.herokuapp.com that is essentially the same as the old app:
- downloads from airtable as before
- serves the data-validation page

### DIFF
--- a/server.js
+++ b/server.js
@@ -54,15 +54,6 @@ const copyValidTables = (oldData, newData) => {
 Promise.resolve(getAllData()).then(allData => {
   let data = allData.airtableData;
 
-  setInterval(function() {
-    Promise.resolve(airTable.hydrateFromAirtable()).then(newData => {
-      copyValidTables(data, newData);
-      Logger.info("Cache refreshed automatically @ " + data.timestamp, {
-        source: "/server.js"
-      });
-    });
-  }, 1000 * 60 * 60);
-
   app.prepare().then(() => {
     const server = express();
     server.use(compression());
--- a/server.js
+++ b/server.js
@@ -54,6 +54,17 @@ const copyValidTables = (oldData, newData) => {
 Promise.resolve(getAllData()).then(allData => {
   let data = allData.airtableData;
 
+  if (staging) {
+    setInterval(function() {
+      Promise.resolve(airTable.hydrateFromAirtable()).then(newData => {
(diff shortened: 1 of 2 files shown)


server.js
```diff
@@ -36,15 +54,6 @@ const copyValidTables = (oldData, newData) => {
 Promise.resolve(getAllData()).then(allData => {
   let data = allData.airtableData;
 
-  setInterval(function() {
```
smcmurtry, line 39: Can we still have this behaviour on staging? That way when we have a permanent staging environment it will be less likely to be out of sync with airtable.

sastels: huh, yeah, I figured it wasn't needed anymore, but we can keep it. Just need to ensure that business owners know they have to refresh manually before getting their changes.

server.js
```diff
@@ -4,10 +4,12 @@ const next = require("next");
 const Cookies = require("universal-cookie");
 const helmet = require("helmet");
 const compression = require("compression");
+const fs = require("fs");
 
 const { parseUserAgent } = require("detect-browser");
 
 const dev = process.env.NODE_ENV !== "production";
+const staging = process.env.STAGING !== undefined;
```
smcmurtry, line 12: don't you also have to add this to .circleci/config.yml and Dockerfile?

sastels: When we deploy the staging version to Azure we can just use the dockerfile we're already using, and inject the STAGING variable through Azure.

Wait, but why didn't I need to add STAGING to the app.json file that Heroku uses? 🤔

sastels: well, the review app says
```
2019-02-06T18:44:41.333291+00:00 app[web.1]: Production environment, using static file
```
while the staging app says
```
2019-02-06T18:14:21.062285+00:00 app[web.1]: Staging environment, downloading from Airtable
```
🤷‍♂️

smcmurtry (approved):
Looks good to me! Just a minor comment about env variables, and can you run `yarn download` again before you merge? We've had some changes to next steps.

PR #1819, opened 2019-02-07
Test json data

### LINKED ISSUE #1805: Test Airtable flat files
once #1802 merges, write tests for the flat files to catch content change errors (possibly run the data validation tests?)

sastels:
resolves #1805 

Perform the tests that are on the data validation page on the static Airtable data (via `yarn test` and as part of CI).

The exception is the testing of the urls. I don't think they should be part of CI, as then CI could fail because of network flakiness. Instead I added a `yarn test_urls` function to test the urls manually.

Note that there's one benefit, (Peer Support) that currently fails the data-validation tests since it doesn't have a url. The corresponding lines in the jest tests are commented out in this PR.

### DIFF
--- a/__tests__/data/area_offices_data_test.js
+++ b/__tests__/data/area_offices_data_test.js
@@ -14,17 +14,11 @@ describe("Area offices data", () => {
     data.forEach(a => {
       expect(
         !!a.address_en &&
-          a.address_en !== "" &&
           !!a.address_fr &&
-          a.address_fr !== "" &&
           !!a.lat &&
-          a.lat !== "" &&
           !!a.lng &&
-          a.lng !== "" &&
           !!a.name_en &&
-          a.name_en !== "" &&
-          !!a.name_fr &&
-          a.name_fr !== ""
+          !!a.name_fr
       ).toBeTruthy();
     });
   });
--- a/__tests__/data/area_offices_data_test.js
+++ b/__tests__/data/area_offices_data_test.js
@@ -1,0 +1,12 @@
+import fs from "fs";
+
+describe("Area offices data", () => {
+  let data;
+  beforeEach(() => {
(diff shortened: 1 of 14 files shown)


__tests__/data/area_offices_data_test.js
```diff
@@ -0,0 +1,31 @@
...
+  });
+
+  it("is not empty", () => {
+    expect(data).not.toHaveLength(0);
+  });
+
+  it("has required fields in all rows", () => {
+    data.forEach(a => {
+      expect(
+        !!a.address_en &&
```
sastels: This `!!` turns the string into a boolean. Without it when you && everything together you end up with a string rather than a boolean.

utils/url_check.js
```diff
@@ -28,13 +29,13 @@ var checkURL = (exports.checkURL = async function checkURL(
   }
 });
 
-var fetchUrl = async function fetchURL(url) {
+var fetchUrl = (exports.fetchUrl = async function fetchURL(url) {
   return await fetch(url).then(
     resp => {
-      return resp.url !== url ? false : true;
+      return resp.url === url;
```
sastels, line 35: This just simplifies the logic here, my IDE kept complaining. 😜

smcmurtry (approved):
Looks good! Every time I run `yarn test_urls` locally I get a different result, maybe it's my bad wifi?

sastels:
> Looks good! Every time I run `yarn test_urls` locally I get a different result, maybe it's my bad wifi?

I think that if you run it quickly in succession something somewhere starts throttling things? Maybe? A bit odd, granted. Let me add a sleep() into things and see if it helps...

sastels:
I added a 20 ms pause between the url fetches, it seems to consistantly work now.

Issue #1817, opened 2019-02-06
Add/fix styling on feedback page


SupeDeDupe:
Font colours should be greyishBrown, and need to fix some padding/margins.

Looks like: 
![screen shot 2019-02-06 at 2 20 37 pm](https://user-images.githubusercontent.com/17673146/52367574-68f65080-2a1a-11e9-9852-fd5ca21f5ef4.png)

Should look like:
![screen shot 2019-02-06 at 2 20 44 pm](https://user-images.githubusercontent.com/17673146/52367586-6e539b00-2a1a-11e9-983b-8cac237703c3.png)

Design link: https://app.zeplin.io/project/5acbd659dc553ec84805afd4/screen/5c3e0c5acd24f8525cbde78c

SupeDeDupe:
![](https://github.trello.services/images/mini-trello-icon.png) [Users can provide feedback on the tool](https://trello.com/c/AHCOa5Nw/104-users-can-provide-feedback-on-the-tool)

SupeDeDupe:
Closed by #1823

PR #1823, opened 2019-02-08
Created feedback_submitted pg + fixed feedback pg styling

### LINKED ISSUE #1821: Add feedback page response page
screen shot 2019-02-08 at 10 20 17 am

https://app.zeplin.io/project/5acbd659dc553ec84805afd4/screen/5c5b2b6a91c96d430b24369d

SupeDeDupe:
Closes #1821 and #1817

### DIFF
--- a/components/radio_selector.js
+++ b/components/radio_selector.js
@@ -13,6 +13,12 @@ const formControl = css`
 `;
 const formLabel = css`
   margin-bottom: 10px;
+  color: ${globalTheme.colour.greyishBrown};
+  font-size: 20px;
+  font-weight: normal;
+  font-style: normal;
+  font-stretch: normal;
+  letter-spacing: normal;
 `;
 const radioOption = css`
   margin-bottom: 10px;
--- a/components/text_area.js
+++ b/components/text_area.js
@@ -12,20 +12,20 @@ const HintText = styled("span")({
 });
 
 const LabelText = styled("span")({
-  fontFamily: globalTheme.fontFamily,
+  fontFamily: `${globalTheme.fontFamily}`,
   WebkitFontSmoothing: "antialiased",
   MozOsxFontSmoothing: "grayscale",
   display: "block",
   clear: "none",
   fontWeight: 400,
-  fontSize: "18px",
+  fontSize: "20px",
   lineHeight: "1.5",
   color: `${globalTheme.colour.greyishBrown}`,
   paddingBottom: "5px"
 });
 
 const FooterText = styled("span")({
-  fontFamily: globalTheme.fontFamily,
+  fontFamily: `${globalTheme.fontFamily}`,
   WebkitFontSmoothing: "antialiased",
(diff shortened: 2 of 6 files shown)


SupeDeDupe:
Need to create tests. Also, I accidentally committed to master, but I think I successfully undid that...

components/radio_selector.js
```diff
@@ -13,6 +13,12 @@ const formControl = css`
 `;
 const formLabel = css`
   margin-bottom: 10px;
+  color: ${globalTheme.colour.greyishBrown};
```
SupeDeDupe, line 16: The radio button stying wasn't following design so much before...

components/text_area.js
```diff
@@ -12,20 +12,20 @@ const HintText = styled("span")({
 });
 
 const LabelText = styled("span")({
-  fontFamily: globalTheme.fontFamily,
+  fontFamily: `${globalTheme.fontFamily}`,
```
SupeDeDupe: I think this is the notation?? I must not have noticed the way I was writing it the first time in this file.

server.js
```diff
@@ -124,10 +124,10 @@ Promise.resolve(getAllData()).then(allData => {
             source: "/server.js"
           });
         });
-      } else if (req.url.includes("data-validation") && !staging) {
-        res
-          .status(404)
-          .send("The Data Validation page only exists on the staging app.");
+        // } else if (req.url.includes("data-validation") && !staging) {
```
SupeDeDupe: Commenting this out allows devs to refresh cache on local host to see the airtable changes during development. But this line should never be commented out in the master branch.

SupeDeDupe (commented):
Data will need to be pulled from Airtable at the same time this PR is merged so we don't have a "key" showing up everywhere, like here:

https://vac-poc-pr-base-pr-1823.herokuapp.com/feedback_submitted

This is something we'll have to think about when adding content in Airtable for features being developed --> the conflict being that there could be admin users testing things in Airtable + the staging app which aren't ready to be pushed to prod, but the devs have a feature + content ready.

server.js
sastels: We could have some sort of "DEV" environment variable flag on our laptops that would cause the data-validation page to be there. Possibly something like the app would use the static `data/data.json` file, but you could go to the data-validation page and click Refresh if you wanted to load from live Airtable.

SupeDeDupe: Yeah, that could work

components/text_area.js
```diff
@@ -18,9 +18,9 @@ const LabelText = styled("span")({
   display: "block",
   clear: "none",
   fontWeight: 400,
-  fontSize: "18px",
+  fontSize: "20px",
```
smcmurtry, line 21: in the design this is 24px

smcmurtry: Also the multiple choice question is 24px and the options are 21px

smcmurtry (approved):
Looks good! Ignore my font size comments. 🦑

PR #1824, opened 2019-02-11
Removed unused components

### LINKED ISSUE #1812: Delete SelectionsEditor, ProfileSelector, NeedsSelector
And associated tests. We're no longer using these components.

brdunfield:
Closes #1812


SupeDeDupe:
I don't think it makes sense to delete the SelectionsEditor at this time since there's still a lot of discussion on possibly bringing it back, depending on how testing goes.

smcmurtry:
@SupeDeDupe It is pretty easy to checkout an older version of master where the components exist and copy them from there. But I'm good either way.

SupeDeDupe (approved):
Alright, shouldn't be too bad to find again if needed, but we might as well remove it now then, in case we don't end up adding it back and then later forget to remove it.

PR #1825, opened 2019-02-11
Updating data.json content from airtable

SupeDeDupe:
1st attempt at updating the content using this new method.

Scott and Allison agreed changes on their end are ready to be merged, and new data from the translation sheet is ready to be used in the feedback_submitted page.

### DIFF
--- a/data/data.json
+++ b/data/data.json
@@ -1225,6 +1225,8 @@
       "oneLineDescriptionEn": "A clinic that assesses and treats mental health issues related to your service.",
       "oneLineDescriptionFr": "Clinique externe qui évalue et traite tout problème de santé mentale causé par votre service.",
       "sortingPriority": "medium",
+      "noteEn": "You need a referral from a Case Manager to visit the clinic.",
+      "noteFr": "(fra) You need a referral from a Case Manager to visit the clinic.",
       "benefitExamples": [
         "recHaX4XKETPQKhHp",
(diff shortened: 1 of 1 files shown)


data/data.json
```diff
@@ -4081,10 +4089,28 @@
...
       "Safe to Delete?": []
+    },
+    {
+      "section": "feedback",
+      "key": "submitted",
+      "English": "Thank you for taking the time to give us feedback. We are listening and constantly improving.",
+      "French": "Merci d’avoir pris le temps de nous faire part de vos commentaires. Nous sommes à l’écoute et nous nous améliorons constamment.",
+      "id": "feedback.submitted",
+      "Used by code": [],
+      "Safe to Delete?": []
```
SupeDeDupe, line 4083: We can probably delete the "Used by code" and "Safe to Delete" columns in translations in Airtable.

sastels: But is it safe to delete the "Safe to Delete" column? 🤔

data/data.json
```diff
@@ -5383,5 +5339,5 @@
     }
   ],
   "errors": [],
-  "timestamp": 1549485129335
+  "timestamp": 1549907642205
```
sastels, line 5342: huh, at some point we should probably save the timestamp as a human readable string

sastels (approved):
LGTM!

PR #1827, opened 2019-02-11
Writing beta feedback to airtable

### LINKED ISSUE #1797: Create a space in Airtable to store data from "Feedback" page
Data from feedback page should be written into Airtable when the user sends their response. We'll probably need fields to hold data for each question on the feedback page ("Overall, how was your experience with this online tool," and "Tell us in more detail what you think about this online tool."): 

https://app.zeplin.io/project/5acbd659dc553ec84805afd4/screen/5c3e0c5acd24f8525cbde78c

https://airtable.com/tblB3Kh0jN8KFRsmW/viwK38zsM3QsI6HWM is already similar.

### LINKED ISSUE #1798: Add writing functionality to "feedback" page
Once #1793 is closed, add functionality of collecting the data and writing it into Airtable. This can be worked on in sync with #1797.

smcmurtry:
Closes #1797, closes #1798

### DIFF
--- a/pages/feedback_submitted.js
+++ b/pages/feedback_submitted.js
@@ -13,10 +13,6 @@ const padding = css`
   padding-top: 15px;
   padding-bottom: 15px;
 `;
-const prevButton = css`
-  margin-top: 50px;
-  margin-bottom: 15px;
-`;
 const textStyle = css`
   fontfamily: ${globalTheme.fontFamily};
   color: ${globalTheme.colour.greyishBrown};
@@ -43,10 +39,6 @@ export class FeedbackSubmitted extends Component {
         backgroundColor={globalTheme.colour.white}
       >
         <Container className={padding} id="mainContent">
-          <HeaderLink href={"/feedback"} className={prevButton} arrow="back">
-            {t("back")}
-          </HeaderLink>
-
           <Header headingLevel="h1" size="lg">
             {t("feedback.page_header")}
           </Header>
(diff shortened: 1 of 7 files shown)


pages/feedback_submitted.js
```diff
@@ -43,10 +39,6 @@ export class FeedbackSubmitted extends Component {
         backgroundColor={globalTheme.colour.white}
       >
         <Container className={padding} id="mainContent">
-          <HeaderLink href={"/feedback"} className={prevButton} arrow="back">
```
smcmurtry, line 46: removed following discussion w/ Michelle since the user has already submitted feedback, so they shouldn’t need to get back to that page

utils/airtable_es2015.js
```diff
@@ -161,7 +161,23 @@ var hydrateFromAirtable = (exports.hydrateFromAirtable = async function hydrateF
 var writeFeedback = (exports.writeFeedback = async function writeFeedback(
   payload
 ) {
-  var url = "https://api.airtable.com/v0/appoFDwVvNMRSaO6o/feedback";
```
smcmurtry, line 164: found a bug! the base key should not have been hardcoded

SupeDeDupe: ok isn't this a security a security issue now, if the key is saved on git?

smcmurtry: It's not a really a key in the sense that knowing it gives you access to anything. It's just a unique id for the airtable base we're using. The read/write keys are what allow you to access the base.

SupeDeDupe (approved):
Nice! 🦑

PR #1835, opened 2019-02-13
Cypress

sastels:
Use Cypress for E2E testing. Here have added a few tests around the guided experience code.

Running on CI has been stable, since I fixed some things that were making it flaky.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -12,14 +12,16 @@ jobs:
       # Download and cache dependencies
       - restore_cache:
           keys:
-            - v2-dependencies-{{ checksum "package.json" }}
+            - v3-dependencies-{{ checksum "package.json" }}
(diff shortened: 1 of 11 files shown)


smcmurtry:
Very cool!

I tried running `yarn cypress run` with my dev server running and the tests failed. But with the production server running the tests passed. Do you know why this is? It would be kind of slow if we have to do `yarn build` and `yarn start` every time while debugging if a test fails.

sastels:
yeah, it doesn't seem to work with the dev server. webpack issues or something :(

**hopefully** if the cypress tests fail then the problem is obvious - the failure is something like "can't find button ... to click on" - and you can go and ensure that you didn't change the id cypress is looking for.

smcmurtry (approved):
Nice work! 🦑

PR #1839, opened 2019-02-13
Update beta banner

### LINKED ISSUE #1832: Update beta banner to include link to feedback page
Also we will need to update the airtable data to get the new strings.

smcmurtry:
Closes #1832 

I added the url prop to `Layout` so that we can link to the feedback page and keep the existing url query parameters so we don't lose the user's answers.


brdunfield (approved):
🦑 Looks good!

PR #1840, opened 2019-02-13
Using the new Airtable data in tests

### LINKED ISSUE #1804: use airtable flat files in testing components
once #1802 merges, use the files in tests (instead of mocks) where appropriate

smcmurtry:
Closes #1804 

I replaced all the data coming from the `_complex` fixtures with data directly from airtable. And it didn't actually break that many tests! We can continue to replace the other mocked data - I'll make another issue for that.


brdunfield:
I'm really not sure how I feel about this. I'm not sure that it makes sense for tests to break whenever real-world data changes. Tests should ensure that functions / components work correctly. Having tests break when content changes means more dev time around those content updates.

For example: Tests which check that they return x number of benefits when a test filter is applied (say organizations). If a new organization benefit is added to airtable, all those tests break and need to be fixed. To me this is unnecessary since the function / code has not changed.

smcmurtry:
@brdunfield Definitely a good point. I've changed the selector tests to be less likely to fail if the data changes. It was a bit more work but they are probably better tests now. Or maybe I just re-implemented the functionality I'm testing in the tests... who's to say? 😄 

One advantage we get with testing using the real data is that, when content changes come in, testing that the content directly could help us catch content errors before they go in production. But it probably will be more work to do this. 

What do you think @sastels?

sastels:
I'm kinda on the fence... I guess I figure it's good to have the test mocks pretty static. Though on the other side there's the added work of deciding what should be in the mocks and how / when to update them :confused: 
And like you say, it gives one more line of defence against people making errors in the content.

Though you're not talking about replacing **all** test data with the mocks, right? just where we used the complex fixtures. So maybe that's an ok compromise.

One thing, though, is that any tests for bad content / data that you think up should be explicitly added to the `__tests__/data` section rather than just relying on the other tests to implicitly catch them (ie what if down the road the decision was made to go back to static complex mocks)

smcmurtry:
So I merged some content updates into this branch earlier this week and it broke the tests. I think that is good evidence that @brdunfield is right and this will cause us to have to spend a lot of time fixing tests when we just want to update content. This could be fixed by rewriting lots of tests but that is probably not a great use of our time. I'll close this for now.

PR #1848, opened 2019-02-15
search box highlight fixed

### LINKED ISSUE #1763: Search box turns yellow when keyword is inputted
When I searched a word, the text box background turned yellow.

apmcguigan:
closes #1763

### DIFF
--- a/components/search_box.js
+++ b/components/search_box.js
@@ -46,7 +46,7 @@ const ClearButton = styled("button")({
   display: "block",
   color: globalTheme.colour.cerulean,
   position: "absolute",
-  left: "-50px",
+  left: "-45px",
   padding: "10px",
   width: "45px",
   height: "44px",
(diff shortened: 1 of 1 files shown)


apmcguigan:
Pushing this up as an alternative, overall it looks better but breaks the outline when the search bar is selected. Likely due to the ClearButton covering the outline

search-45px

smcmurtry:
Hmm, not sure this is an improvement. What about removing the yellow background entirely and keeping the focus border? https://stackoverflow.com/questions/2781549/removing-input-background-colour-for-chrome-autocomplete

apmcguigan:
Heroku deployment doesnt seem to be working, currently looks like this on my end in Chrome
search-good

SupeDeDupe (approved):
Nice! 🦑👍

PR #1861, opened 2019-02-19
Content update for new ui text needed for layout changes

smcmurtry:
Looks like this also includes some new translations see more content.


brdunfield (approved):
👍

PR #1864, opened 2019-02-19
added language prop to example bullets component to trigger re render when lang changes

### LINKED ISSUE #1860: See more example content not changing language
To reproduce:
1. go to: https://benefits-prestations.veterans.gc.ca/benefits-directory?lng=en
2. click see more on rehab services
3. switch to french
4. some of the see more content will still be in english

I think the component that displays that content isn't picking up the language change - maybe if we pass it in the language as it's own prop that will fix it? I think this bug is showing up now because the list order is no longer changing between french and english so a full re render isn't happening at the top level.

smcmurtry:
Closes #1860


SupeDeDupe (approved):
🦑 👍

PR #1852, opened 2019-02-18
New directory layout

smcmurtry:
Closes #1836 

Now ready for review. I'm not planning to update link colours/styles as part of this PR. Here's what I did do:

-changed layout on both directory and favourites page to match the new design
-split out some components (`StickyHeader` and `QuickLinks`) so they could be re-used on both pages
-added tests for the new components
-modified the `ShareBox` component to conform to the new design
-removed `ContactUs` component since this is no longer in the designs (it will eventually live in the footer I think?)

Many of the font sizes are not the same as the designs - here I opted to go for the closest existing font size supported by a component. This was because I didn't want to deal with the implications on other parts of the app using the same components as part of this PR since it's big enough already.

PR #1869, opened 2019-02-20
Content update for ui text in new layout

smcmurtry:
This should be merged before #1852 

In addition to UI text this includes some changes to links made by Scott.


brdunfield (approved):
👍

PR #1872, opened 2019-02-21
Content update

SupeDeDupe:


data/data.json
```diff
@@ -1722,6 +1722,27 @@
       "benefitExamples": [],
       "seeMoreSentenceEn": "",
       "seeMoreSentenceFr": ""
+    },
+    {
+      "id": "recU6FYqNjRupolvb",
```
smcmurtry: this looks like an empty row in the benefits table - maybe this is why the tests failed?

SupeDeDupe: yeah that's why it failed the first time

SupeDeDupe:
No longer needed

Issue #1871, opened 2019-02-21
Implement VAC's new header design in the app


smcmurtry:
See: https://www.veterans.gc.ca/fra

smcmurtry:
Closed by #1898

PR #1898, opened 2019-02-25
Cloning the VAC header

smcmurtry:
Working on #1871.

So far I've copy/pasted the html from the VAC site into a new component and converted to JSX by changing `class` to `className`, etc. Then I pasted their css resources into _document.js and converted to JSX by changing `<link ...>` to `<link ... />`. The header looks mostly like theirs but some of their css is impacting our icon styles, so I'll figure that out.

I'm putting this up before it's done so I can test using the review app to see the effect on page size and performance.

Edit: page size went from 400KB to 1MB!! But maybe this is not a problem since most users will click to it from the VAC home page, so they will have already downloaded the massive css files.


smcmurtry:
I think I'll try a different approach

Issue #1904, opened 2019-02-25
Remove "Closest VAC office" from print page


courtneycl:
We don't have this feature anymore. 

![image](https://user-images.githubusercontent.com/3474250/53362380-018c3c00-3908-11e9-9a1c-63e7804c5915.png)

I'll add a trello card to consider updates to the print page (including new next steps, etc).

SupeDeDupe:
Closed by #1909

PR #1909, opened 2019-02-25
Remove map page and area office info

### LINKED ISSUE #1905: Remove Map page
and associated dependencies

brdunfield:
Closes #1905 and #1904


smcmurtry:
Wow that was a big chunk of the app! I checked the network traffic using chrome devtools and it looks like your review app is still downloading some google map resources. They're being added in pages/_document.js.

smcmurtry (approved):
Nice work! 🦑

PR #1911, opened 2019-02-25
removed code that updated url after changing a filter on the directory

### LINKED ISSUE #1900: Clicking 'back' in your browser after landing on results page clears the categories
Repro: 
1. Go through the guided experience, fill out all the q's and select a few categories. (I did Veteran, CAF, yes health issue, emerg funds, career/jobs, education and training) 
2. observe 7 results match my selections: 
image
3. hit the back button in your browser 
4. observe the page reloads and now we have 26 results; the URL removed our categories: 
image
https://benefits-prestations.veterans.gc.ca/benefits-directory?lng=en&patronType=veteran&serviceType=CAF&serviceHealthIssue=hasServiceHealthIssue

Keep hitting back, the browser starts to chop off the rest of the selections and reload the page on the directory. Expected that when you hit back, you would be taken back to the previous page you were on.

smcmurtry:
Closes #1900. Code was no longer needed since we removed the radio buttons from the directory page.

### DIFF
--- a/__tests__/pages/benefits_directory_test.js
+++ b/__tests__/pages/benefits_directory_test.js
@@ -129,15 +129,4 @@ describe("BenefitsDirectory", () => {
     mounted.unmount();
     expect(window.removeEventListener).toBeCalled();
   });
-
-  it("has a correct setURL function", () => {
-    props.profileFilters = { patronType: "veteran", service: undefined };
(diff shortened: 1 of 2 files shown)


courtneycl (approved):
good from a product perspective

PR #1919, opened 2019-02-26
Fixed saved list print pg bug

### LINKED ISSUE #1901: print option from saved list prints out the directory, not the saved list
1. Have a couple things saved and go to your saved list - I have 2:
image
2. Hit print
3. Observe that it will print all results from the directory: 
image

Expected it would just print from our saved list.

SupeDeDupe:
Closes #1901


smcmurtry (approved):
Looks good! 🦑

PR #1929, opened 2019-02-27
Added the utm bit to the share url

### LINKED ISSUE #1924: add utm_source query parameter to the share link
This is so we can track usage of the share feature. Just need to add `utm_source=share` or something like that.

SupeDeDupe:
Closes #1924

### DIFF
--- a/components/share_modal.js
+++ b/components/share_modal.js
@@ -185,7 +185,9 @@ class ShareModal extends Component {
         type="text"
         id={shareTargetId}
         defaultValue={
-          this.state.origin + this.props.url.asPath + this.state.utm
+          this.props.url.asPath.includes(this.state.utm)
+            ? this.state.origin + this.props.url.asPath
+            : this.state.origin + this.props.url.asPath + this.state.utm
         }
         contentEditable="true"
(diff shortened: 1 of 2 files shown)


smcmurtry:
Hmm, I think the url isn't quite right. If you click share from here: https://vac-poc-pr-base-pr-1929.herokuapp.com/benefits-directory

This is the share url: https://vac-poc-pr-base-pr-1929.herokuapp.com/benefits-directory/utm_source=share-link&utm_medium=fbas

which doesn't work. Instead of `/utm_source...` it should be `?utm_source...`.

smcmurtry:
So now if I start from here: https://vac-poc-pr-base-pr-1929.herokuapp.com/benefits-directory?lng=en&patronType=veteran

I get this share link: https://vac-poc-pr-base-pr-1929.herokuapp.com/benefits-directory?lng=en&patronType=veteran?utm_source=share-link&utm_medium=fbas

Which works but has a syntax error - `...veteran?utm_source...` should be `...veteran&utm_source...`.

smcmurtry (approved):
Looks good! 🦑

PR #1930, opened 2019-02-27
Fix for share copy link function only working sometimes

### LINKED ISSUE #1902: Copy link functionality only works sometimes
Not sure what the case is for a repro; try refreshing the page and trying it. I can get it to work as expected ~50% of the time. Other times I can click the button but do not receive `Link copied`, and the link does not copy to my clipboard. 

https://benefits-prestations.veterans.gc.ca/benefits-directory?lng=en&patronType=veteran&serviceType=CAF&serviceHealthIssue=hasServiceHealthIssue&selectedNeeds=recHB4dscib9fjxso,recIoUeSmWb2pyYaD

brdunfield:
Closes #1902 

see: https://stackoverflow.com/questions/37270787/uncaught-syntaxerror-failed-to-execute-queryselector-on-document

### DIFF
--- a/components/share_modal.js
+++ b/components/share_modal.js
@@ -128,7 +128,9 @@ class ShareModal extends Component {
 
   copyText(e) {
     let t = e.target.dataset.copytarget;
-    let shareInput = t ? document.querySelector(t) : null;
+    let shareInput = t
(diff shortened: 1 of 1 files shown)


smcmurtry (approved):
Nice fix! 🦑

PR #1935, opened 2019-02-27
New header

smcmurtry:
Closes #1898 

This doesn't include the menu (I'll make a separate issue for that). I know this isn't a perfect match - especially the breakpoints. But those would probably be better changed as part of another PR that makes the overall layout (page width, margins, etc) more consistent with the VAC site.


components/layout.js
```diff
@@ -21,8 +21,10 @@ const Content = styled("div")`
   min-height: calc(100vh - 65px);
 `;
 const header = css`
-  background-color: ${globalTheme.colour.greyishBrownTwo};
+  background-color: ${globalTheme.colour.bannerColour};
   padding: 0px;
+  // display: flex;
+  // justify-content: center;
```
SupeDeDupe: is this still needed?

components/myvac_button.js
```diff
@@ -0,0 +1,39 @@
...
+    padding-left: 10px;
+    color: ${globalTheme.colour.blueGrey};
+  }
+`;
+
+class MyVacButton extends Component {
+  render() {
+    const { t } = this.props;
+    return (
+      <FooterLink href={t("myvac_register_href")} className={desktopButton}>
```
SupeDeDupe: Assuming most of our users are new to VAC, it can make sense to use the register link here, but are we doing this on purpose? The VAC site header uses https://www.veterans.gc.ca/eng/e_services/ for the My VAC Account button.

components/federal_banner.js
```diff
@@ -40,52 +23,80 @@ const container = css`
...
+    const { t, skipLink, i18n, url } = this.props;
     return (
-      <div>
-        <SkipToMainContent skipLink={skipLink} t={t} />
-        <div className={container}>
-          <div className="svg-container">
-            <FIP fillColor="white" t={t} />
+      <Grid container>
+        <Grid item xs={12} md={6}>
+          <SkipToMainContent skipLink={skipLink} t={t} />
```
SupeDeDupe, line 75: Just a note for the future PR: getting the styling just right to match the VAC site header with the breakpoints might be a bit tricky to do anyway, because of the space the SkipLink takes up.

SupeDeDupe (commented):
Awesome! Just a few comments...

components/myvac_button.js
smcmurtry: Good catch! I didn't use a different link on purpose.

SupeDeDupe (approved):
Good job! 🥇👍 I already merged the content update. Was going to merge that into this branch so I could check the link to be connected properly, but ran into something weird and didn't want to break anything, so I left it for you.

PR #1943, opened 2019-03-01
Remove pagewidth var

### LINKED ISSUE #1912: Remove the page width variable from redux
Also the listeners that update this variable on the directory page and probably in other places. This is no longer needed or used (AFAIK).

brdunfield:
Closes #1912


smcmurtry (approved):
Nice work! 🦑

PR #1950, opened 2019-03-04
VAC footer

### LINKED ISSUE #1940: Implement VAC's footer as a react component
We can base this on the version on their live site: https://www.veterans.gc.ca/eng

smcmurtry:
Closes #1940

Instead of putting all the text in airtable, I made a separate english and french version of the footer. This is because I don't think anyone on our team will be changing this content. What is more likely is that the content would change on the main VAC website and we would update our component to reflect this.

To create the components I just clicked "View Page Source" in chrome, pasted in the footer html, and changed the syntax to JSX (class => className, adding slashes, etc). To get the css (in vac_css_used.css) I used the chrome extension [CSS Used](https://chrome.google.com/webstore/detail/css-used/cdopjfddjlonogibjahpnmjpoangjfff?hl=en) to extract the minimum necessary css to reproduce the styling.

The only issue with the styling is that one of the font types doesn't match - I think this must be because we don't have the same default font set as veterans.gc.ca. Maybe that could be fixed as part of another PR.


smcmurtry:
Actually I just noticed this is effecting our svg styling - I'll try to fix that.

smcmurtry:
Okay I was able to use [LESS](http://lesscss.org/less-preview/) to scope the VAC css file to a div containing the footer, so that fixes the styling collisions. All I did was pasted in my extracted css surrounded by another selector: `#footer_styles {...extracted css}`  into the tool above, and then pasted the output back to the vac_css_used.css file.

brdunfield:
There's different text in the VAC Assistance Service section - maybe updated after you started working on it?

Padding seems a bit different too, particularly on the list of links at the left.
We're also missing the image background for the very bottom set of links - are we planning to add that or just skip it?

smcmurtry:
I updated the content in the VAC Assistance Service section. The leaf background image at the bottom was 252KB which almost doubled our page weight so I left it out for now. 

I'll try to figure out what's causing the difference in padding.

smcmurtry:
Okay I was able to fix the list padding and the font. Now I think the only styling that's wrong is the share button spacing, but I'm not sure what's happening there so maybe that can be another PR.

brdunfield (approved):
Works for me, I made an issue for that.

I think my only other comment is that the css stripped from the VAC site copies a few things, mostly re-importing the same fonts we're already using. But I'm not sure if this is worth manually changing, as I assume they'll come back whenever updates are made using this process.

Great work! 🦑

PR #1951, opened 2019-03-05
Content update

SupeDeDupe:
Scott requested this a content update to ensure Transition interview is invisible in the app.

FYI, maybe it's worth reevaluating some tests. The cypress tests failed in the first push because "Select who will be receiving the benefits" was hardcoded in several places. Updating to "Select who would be receiving the benefits" fixed them. Should these lines be hardcoded?


smcmurtry (approved):
Looks good! You could make the cypress tests get the text directly from the data.json file. But that would probably make it less readable.

PR #1956, opened 2019-03-06
Improve contrast of text

### LINKED ISSUE #1810: WCAG 2.1 AAA contrast ratio
In keeping with striving for WCAG 2.1 AAA this is something we can look at.

**1.4.6 Contrast (Enhanced)**
The color of the text and the color of the background are not in sufficient contrast to each other.

The contrast ratio should be at least 7:1 for normal text and 4.5:1 for large text.

Note: Contrasts are not relevant where content is hidden. But make sure that for hidden content, which becomes visible when receiving focus, the contrast ratio is sufficient.

Insufficient contrast between text and its background can give problems for users with visual impairments and color blindness, but it can also affect a lot of other users, leading to important text potentially being overlooked.

**How do I meet this criteria?**
G17 Ensuring that a contrast ratio of at least 7:1 exists between text (and images of text) and background behind the text
G18 Ensuring that a contrast ratio of at least 4.5:1 exists between text (and images of text) and background behind the text

screen shot 2019-02-05 at 8 09 28 pm

brdunfield:
Closes #1810 

Designs on Zepplin have been updated with the new text colour used in this PR. I have verified with the Axe browser extension that these contrast ratios are acceptable.

### DIFF
--- a/components/header_button.js
+++ b/components/header_button.js
@@ -11,7 +11,7 @@ const style = css`
   font-size: 24px;
   line-height: 44px;
   font-weight: bold;
-  color: ${globalTheme.colour.slateGrey};
+  color: ${globalTheme.colour.charcoalGrey};
   background-color: transparent;
   border: none;
   text-decoration: none;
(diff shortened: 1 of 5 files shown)


brdunfield:
[Accessibility issues](https://trello.com/c/knKtuM4T/100-accessibility-issues)

smcmurtry (approved):
I tested with the axe browser extension on my end too and I'm no longer getting any colour contrast issues on any page. Great work!

PR #1971, opened 2019-03-08
Security patch

SupeDeDupe:
Additional dependency updates for #1957. The security warnings on some of the dependencies still exist. I think this will solve the issue.


thomascraig (approved):
nice, fixed merge@ version

PR #1973, opened 2019-03-08
ARIA Landmark Audit

### LINKED ISSUE #1757: Content not included in landmarks
From Julianna:

Found with SiteImprove

Content not included in landmarks
1.3.1 Info and Relationships
HTML5 or WAI-ARIA landmarks are used on the page, but not all content is included.

When using HTML5 or WAI-ARIA landmarks it is best practice to include all content on the page in landmarks. In this way users of assistive technologies can use the landmarks for navigating the page without losing track of content.

Make sure that all content on the page is included in HTML5 or WAI-ARIA landmarks.

*Code:*

```<noscript style="transition: none 0s ease 0s;"><div class="css-1wjadmq"><div class="css-1ri8owv"><div class="copy"><p>Javascript is disabled and you may experience reduced functionality. To see a full list of benefits, click <a href='/all-benefits'>here</a>.</p></div></div></div></noscript>```

*Notes*: I'd like to discuss potential false positive.

### PR DESCRIPTION
**Pull request by brdunfield**
[WIP]
Resolves #1757 , #1809 , #1757 

The VAC footer contains two `<nav>` elements that do not have distinguishable labels. I'm not sure whether that's something we should address or not, since updates to this component will be coming from their site and override ours. Thoughts?

### CODE REVIEW

**Comment by brdunfield** [comment]
[Accessibility issues](https://trello.com/c/knKtuM4T/100-accessibility-issues)

**Comment by thomascraig** [comment]
@brdunfield Which two labels/elements and I can submit to comms to resolve.

**Comment by brdunfield** [comment]
screen shot 2019-03-08 at 9 37 25 am
screen shot 2019-03-08 at 9 37 35 am

@thomascraig

**Comment by brdunfield** [comment]
Slightly modified the RadioSelector component so that the legend (if it has one) is read out by the screenreader when the radio selector gets focus.
This doesn't work in GE where the "labels" are H2s outside of the radio selector. I'll make an issue for it.

**Comment by brdunfield** [comment]
I'm undo-ing most of my changes to RadioSelector - turns out it was causing failing tests due to multiple labels.

**Comment by smcmurtry** [comment]
My interpretation of #1757 is a bit different. An example of an element on the page that isn't included in an ARIA Landmark would be our feedback bar at the bottom - it's not in `<div role="main">` or `<footer>`. There are probably other things on the page like this too, like lots of unneccessary divs that we may be able to replace with `<React.Fragment>` in some cases.

Also, apparently you can make an ARIA Landmark either way: `<div role="main">` is the same as `<main>` (see this [Stack Overflow discussion](https://stackoverflow.com/questions/19595719/html5-main-element-versus-aria-landmark-role-main)). I think `<main>` is a bit more obvious.

Also, we may be able to use more types of landmarks where it fits our content - [here's a list of all of them from W3C](https://www.w3.org/WAI/GL/wiki/Using_ARIA_landmarks_to_identify_regions_of_a_page)

**Comment by brdunfield** [comment]
I agree there seems to be inconsistency in our use of elements like `<nav>` but not `<main>` . Was any of this done for browser compatibility reasons?

I added the form role to the feedback page. Perhaps it should also be added in the GE pages or the footer feedback form.

File: pages/feedback.js [hunk 1]
```diff
@@ -97,60 +97,61 @@ export class Feedback extends Component {
...
-              arrow={true}
-              size="big"
-              onClick={() => {
-                this.sendFeedback();
-                Router.push({
-                  pathname: "/feedback_submitted",
-                  query: url.query
-                });
-              }}
+          <div role="form">
```
smcmurtry: Probably better to use a `<form>` element here instead. [From Mozilla](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/Complementary_role): "Developers should always prefer using the correct semantic HTML element over using ARIA"

brdunfield: Changed on Feedback page. The other role="form" is being used on a <Grid> element and so probably will need to stay as a role.

smcmurtry (approved):
Looks good 🦑

PR #1976, opened 2019-03-08
Upgraded lodash

smcmurtry:
We had lodash v4.17.11 and v4.17.10 installed because it was a dependency of some of our dependencies (aka a subdependency). You can see which ones by running `yarn why lodash`. You can't really upgrade your subdependencies directly, so I just deleted the yarn.lock file, deleted node_modules/, and re-ran `yarn install` and that caused the latest version of lodash to be installed.


thomascraig (approved):
perfect, thanks.

PR #1986, opened 2019-03-11
Redo header to be closer to vererans.gc.ca

### LINKED ISSUE #1959: Update the top banner to include a link to canada.ca
screen shot 2019-03-06 at 4 37 08 pm

smcmurtry:
Closes #1959


brdunfield:
It looks like VAC's header has a bit more padding on the bottom, as well as slightly different spacing between the icons on the right (and on mobile).

smcmurtry:
@brdunfield I added some padding and I've been trying to debug the styling differences for the last couple hours. I added some missing styles, and as far as I can tell, the computed css for this component is now the same as on veterans.gc.ca. I also added a non breaking space to the My VAC link. It's still not a perfect match but I think this is the best I can do.

brdunfield (approved):
Looking good! 🦑

PR #1995, opened 2019-03-13
Finish Upgrading Emotion

### LINKED ISSUE #1588: finish emotion 10 upgrade
We have partially migrated to emotion 10, following the migration path as described in https://emotion.sh/docs/migrating-to-emotion-10. We have to finish the migration. Essentially, instead of passing the ``css`...` `` using the `className` prop, we should use the `css` prop.

brdunfield:
Resolves #1588 

Getting this warning in the console: "Warning: Style property values shouldn't contain a semicolon."

Will address this in a separate issue / PR


smcmurtry:
Not sure of the cause, but now there's less whitespace between the radio buttons in the GE. Maybe there's a style not getting applied correctly?

smcmurtry (approved):
Great work Bethany! That is a big job done. 💯 🦑

PR #1996, opened 2019-03-13
Minor style inconsistencies

### LINKED ISSUE #1988: Minor style inconsistencies
See: https://www.veterans.gc.ca/eng/health-support/hearing-loss-and-tinnitus/compensation-hearing-loss-tinnitus

Our app's green bar looks different, and we have a different drop shadow behind our Paper components.

apmcguigan:
closes #1988

Everything looks like it matches but I wasn't 100% certain on whether I shouldve deleted tea from theme and same thing with overwriting boxShadowMUI. 

They were both only used on that one page but didn't know if I should keep them just in case.

### DIFF
--- a/components/dropdown.js
+++ b/components/dropdown.js
@@ -39,7 +39,7 @@ const selectStyle = css`
   line-height: normal;
   width: 100%;
   height: 44px;
-  box-shadow: ${globalTheme.boxShadowMui};
+  box-shadow: ${globalTheme.boxShadow};
   border: 0px;
   :focus {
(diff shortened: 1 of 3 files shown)


apmcguigan:
Feel free to merge if approved since I won't be back in until Friday

smcmurtry (approved):
The styling changes look great! My only suggestion would be to rename the `boxShadowMui` to be `boxShadow` or something like that. The `Mui` part means that it is the box shadow from Material UI.

PR #2002, opened 2019-03-14
Made the space at the bottom more consistent

### LINKED ISSUE #1965: Add space below the body content, before the footer in GE
Use the same amount of space that's being used in the directory. It helps to separate the content, especially in the GE where there are questions and it doesn't look like the footer feedback question is part of that process.

smcmurtry:
Closes #1965


SupeDeDupe (approved):
Nice! 🦑

PR #2008, opened 2019-03-18
Cleaned up the all benefits page

### LINKED ISSUE #1541: Update all-benefits.js
This page is used in case the user has javascript disabled. The "List of all Benefits" text seems to be out of date. The save button should be hidden. Maybe the dropdown that shows the benefit children on cards should also be hidden. The "Learn More" needs functionality repaired.

screen shot 2018-12-06 at 1 42 08 pm

smcmurtry:
Closes #1541 

Added some padding around the heading and a max-width to the benefit cards. Did a quick check with Michelle that this was a good idea.

### DIFF
--- a/pages/all-benefits.js
+++ b/pages/all-benefits.js
@@ -1,13 +1,21 @@
-import React, { Component } from "react";
+/** @jsx jsx */
+import { Component } from "react";
+import { css, jsx } from "@emotion/core";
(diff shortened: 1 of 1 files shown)


brdunfield (approved):
Looks good, although this will conflict with my emotion PR - I'll make sure to wait until this is merged and update appropriately.

🦑 💯

PR #2009, opened 2019-03-18
Mock app element to fix warning "App element is not defined" warning

### LINKED ISSUE #1461: share_modal tests warning
warning about App element being undefined in the share_modal tests. This is an artifact of the react-modal dependency component.

smcmurtry:
Closes #1461


brdunfield (approved):
Awesome! This has been so annoying for so long! 🦑 💯

PR #2012, opened 2019-03-18
Fixed feedback submitted bug

### LINKED ISSUE #1999: Feedback confirmation page is missing
1. Be on the feedback page: https://benefits-prestations.veterans.gc.ca/feedback
2. fill out the form, hit `Send`
3. observe the page briefly changes to `/feedback_submitted`, then reloads to an empty feedback page. 

Expected we'd be directed to the feedback confirmation page: 
https://benefits-prestations.veterans.gc.ca/feedback_submitted
https://app.zeplin.io/project/5acbd659dc553ec84805afd4/screen/5c5b2b6a91c96d430b24369d

Note the feedback is captured in Airtable as expected.

smcmurtry:
Closes #1999 

Moving the navigation from the `onClick` event to a `<Link>` component similar to how we do it in `GuidedExperience` seems to fix the problem. Not sure why - maybe the page was re-rendering before the `Router.push` event had a chance to fire?

### DIFF
--- a/__tests__/components/favourite_button_test.js
+++ b/__tests__/components/favourite_button_test.js
@@ -36,13 +36,15 @@ describe("FavouriteButton", () => {
 
   it("is favourited if in list", () => {
     let mounted = mount(<FavouriteButton {...props} />);
-    expect(mounted.find("svg").hasClass("saved")).toEqual(true);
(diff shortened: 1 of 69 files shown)


brdunfield (approved):
Nice! 👍 🦑

PR #2013, opened 2019-03-18
fixed material ui typography warning

### LINKED ISSUE #1994: Material UI Warning
Warning: Material-UI: you are using the deprecated typography variants that will be removed in the next major release.
Please read the migration guide under https://material-ui.com/style/typography#migration-to-typography-v2

smcmurtry:
Closes #1994 

Adding this configuration to our theme fixed the warning, which is to do with new behaviour in material ui (MUI) to map headings to h1, h2, etc elements. I checked and we're not using any of the MUI typography components that would be effected so this should make no difference to our app.


brdunfield (approved):
🙌 🦑

PR #2023, opened 2019-03-19
Disable lunr stemmer

### LINKED ISSUE #1648: Search bugs out midway through words
Search for "disability" letter by letter - it will match words up until about "disabili" at which point search results drop to 0 until you complete the word

brdunfield:
So this fixes #1648 , however it removes the stemmer from lunr. I'm not 100% sure what we've been using this for, so I thought I'd just put this up as a draft PR and get thoughts / insight. @smcmurtry

### DIFF
--- a/store.js
+++ b/store.js
@@ -45,6 +45,8 @@ export const reducer = (state = initialState, action) => {
         this.field("vacNameEn");
         this.field("oneLineDescriptionEn");
         this.field("benefitExamplesEn");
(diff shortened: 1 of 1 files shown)


brdunfield:
Can confirm that just removing the stemmer is not a good idea.
Search for "disabilities" now returns 0 results, whereas in master it still returns 3 (matches against "disability")

brdunfield:
Closing as this is not a good solution to the problem.

PR #2024, opened 2019-03-20
Needs list fixed

### LINKED ISSUE #2021: Needs list getting cutoff
This is the current needs page from Firefox:
Screen Shot 2019-03-19 at 2 46 33 PM

I noticed a while back that in Chrome, this weird rendering happens if a new item is added to the needs list (pic is from Nov 2018):
Screen Shot 2018-11-15 at 11 55 50 AM

Probably the same problem.

apmcguigan:
closes #2021


SupeDeDupe (approved):
Awesome! Thanks for fixing this!! 🙌 🥇

Issue #2028, opened 2019-03-21
Add `aria-label` to Learn More links


smcmurtry:
This is so that a screen reader user can distinguish between the links. Something like:

`aria-label="Learn more about Disability benefits"`.

PR #2037, opened 2019-03-26
Accessibility and bug fix

### LINKED ISSUE #2032: Set role="presentation" on the share icon and the chevron icon
Non text content does not have an accessible name.

* Issue:  The WAI-ARIA image, created with role="img", does not have an accessible name.
* Snippet: `<svg class="svg-inline--fa fa-share-alt fa-w-14 mrgn-lft-sm" aria-hidden="true" focusable="false" data-prefix="fas" data-icon="share-alt" role="img" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 448 512" data-fa-i2svg="" style="z-index: 2147483646; position: relative;">`

How to fix

* If the image is informative or functional, use either the WAI-ARIA attribute 'aria-label' or 'aria-labelledby' to create an alternative text for the image.
* If the image is purely decorative, use the WAI-ARIA attribute role="presentation" instead of role="img" for the image. This will cause assistive technologies to ignore the image.

Screen Shot 2019-03-21 at 11 57 57 AM

smcmurtry:
Closes #2032, #2028, and makes the card expansion icon rotate again.

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,7 +28,7 @@ Contact other developers on the project for what values we're currently using.
 | `WEBHOOK_URL`                | Sends slack deployment notifications                                                                         | production           |
 | `BROWSERSTACK_USERNAME`      | run tests on Windows via BrowserStack                                                                        | locally              |
 | `BROWSERSTACK_ACCESS_KEY`    | run tests on Windows via BrowserStack                                                                        | locally              |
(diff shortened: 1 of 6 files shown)


thomascraig (approved):
Good job

PR #2065, opened 2019-04-04
Update home link in breadcrumbs

### LINKED ISSUE #2010: Clicking "home" breadcrumb should go to veterans.gc.ca/lng
Currently it goes to the 1st page in the GE. This isn't consistent with the way breadcrumbs work on veterans.gc.ca.

SupeDeDupe:
Closes #2010. This PR changes the `Home` breadcrumb href from the guided experience url to the VAC home page url, changes `Find benefits and services` breadcrumb href from the benefits directory page to the first guided experience page, adds a `Results` breadcrumb to replace the previous benefits directory breadcrumb (which used to be `Find benefits and services`), and adds a breadcrumb for the feedback page (`Help us improve`).

It might be worth refactoring the breadcrumbs implementation such that the feedback page breadcrumbs are more flexible. The feedback page is accessible by any page on the site, but the breadcrumbs on the feedback page are currently `Home` > `Find benefits and services` > `Help us improve`. Changing the breadcrumbs to be more dynamic may require keeping track of the current breadcrumb path, without allowing the breadcrumbs to circle back (e.g. `Home` > `Find benefits and services` > `Saved list` > `Help us improve` > `Find benefits and services`. We could store the user's breadcrumbs list in redux, but changes to the breadcrumbs from the user's navigation will require some serious logic to be in place.

An argument for keeping the current breadcrumbs implementation: the user already has the back button option on the feedback page to route to their previous page, and the completion of the the feedback page leads to the feedback submitted page, and then back to the benefits directory. The use cases may require more design input. What do we set the breadcrumbs to in cases such as the one listed above? `Home` > `Find benefits and services` > `Saved list` > `Help us improve` > `Find benefits and services`

 If we decide to change anything about the feedback page breadcrumbs, it can be its own issue.

### DIFF
--- a/__tests__/components/favourites_test.js
+++ b/__tests__/components/favourites_test.js
@@ -49,7 +49,7 @@ describe("Favourites", () => {
       },
       url: { query: {} },
       printUrl: "/print",
-      geUrl: "/",
+      guidedExperienceUrl: "/",
       saveFavourites: jest.fn()
     };
     _shallowFavourites = undefined;
--- a/__tests__/components/favourites_test.js
+++ b/__tests__/components/favourites_test.js
@@ -49,7 +49,7 @@ describe("Favourites", () => {
       },
       url: { query: {} },
       printUrl: "/print",
-      homeUrl: "/",
+      geUrl: "/",
       saveFavourites: jest.fn()
     };
     _shallowFavourites = undefined;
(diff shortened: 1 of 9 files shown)


SupeDeDupe:
@PhilSmith-VAC Thoughts?

PhilSmith-VAC:
I'm cool with leaving the implementation for the feedback page as is.

thomascraig (commented):
Suggest switching naming convention of variables to a more meaningful name:

geUrl => . guidedExperienceURL

thomascraig (approved):
looks good.

PR #2069, opened 2019-04-05
Sidebar bugfix

### LINKED ISSUE #2029: Fix ARIA labels for sidebar
IDs used in ARIA and labels must be unique. 

* Tags: duplicate-id-aria
* Issue: IDs used in ARIA and labels must be unique (duplicate-id-aria)
* Snippet: `<input type="radio" id="WSV (WWII or Korea)-0" class="css-hhofyz" value="WSV (WWII or Korea)">`

How to fix
1. Rename any duplicate ARIA ID values. Duplicate ARIA IDs are common validation errors that may break the accessibility of labels, e.g., form fields, table header cells.
2. To fix the problem, change an ID value if it is used more than once to be sure each is unique.
3. Unique ID's differentiate each element from another and prevent invalid markup, wherein only the first instance gets acted upon by client-side scripting, or where assistive technologies typically only reference the first one accurately.

### LINKED ISSUE #2030: Sidebar form elements missing labels
Form elements must have labels 

* Tags: label
* Issue: IDs used in ARIA and labels must be unique (duplicate-id-aria:)
* Snippet: `<input type="radio" id="veteran-0" class="css-hhofyz" value="veteran">`

How to fix
Form element has multiple <label> elements
Form element has explicit <label> that is hidden

Programmatically associate labels to all form controls and ensure there are no duplicate labels. You can do so by using an implicit `labelelement` and explicit `label`, `aria-label`, or `aria-labelledby` attribute values.

Form elements that should have labels
* Text entry fields, e.g. `<input type="text">`, `<input type="password">` and `<textarea>`
* Radio buttons, `<input type="radio">`
* Checkboxes, `<input type="checkbox">`
* Select menus, `<select>`

The only exceptions for this requirement are:
* Buttons — buttons are self-labeling
* Hidden inputs — Inputs with the type attribute value of hidden (e.g., `type="hidden"`). These inputs are hidden and unavailable for user input. They therefore need no label.

### LINKED ISSUE #2033: Non unique ids because there are 2 sidebar components
The ID used for an element on the page is also used for another element on the page.
Tag: ID-not-unique
Issue: ID's must be unique.
Snippet: `<input type="radio" value="veteran" id="veteran-0" class="css-hhofyz" style="">`

This could be fixed either by removing the mobile sidebar component and making the other component work for both. Or we could change the ids used in one of these components.

### LINKED ISSUE #2062: "Edit selections" sidebar does not appear on mobile
Screen Shot 2019-04-02 at 5 13 58 PM

smcmurtry:
This closes #2062, closes #2033, closes #2030, and closes #2029.

It is a design change - now the open/close arrow is visible/functional on desktop. But it was causing accessibility issues to have 2 different versions of this component (one for mobile, one for desktop) so hopefully this is a good solution.


SupeDeDupe (approved):
Nice job! Thanks for fixing this. 🦑👍

PR #2079, opened 2019-04-12
Improved alignment of need name and tag icon

### LINKED ISSUE #2042: Category icon tags and text don’t line up
On the cards, the category icon tags and text don’t line up right. The text should be centered vertically on the icon and there needs to be space added between the icon and text (~8px).

https://zpl.io/bAqozwB

CategoryTag

smcmurtry:
Closes #2042

### DIFF
--- a/components/benefit_cards.js
+++ b/components/benefit_cards.js
@@ -64,11 +64,10 @@ const floatRight = css`
 `;
 
 const tagStyle = css`
-  width: 20px;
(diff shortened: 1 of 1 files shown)


SupeDeDupe (approved):
Nice job. I removed the width and height because I noticed they were causing weird resizing issues upon reloading the page. Thanks for fixing this! 👍🦑

PR #2080, opened 2019-04-12
Tighten up focus outlines

smcmurtry:
Here is the related issue that this closes: https://github.com/cds-snc/find-benefits-and-services/issues/64


SupeDeDupe (approved):
Looks good to me! 🦑

PR #2081, opened 2019-04-12
Fixed server error when a table is empty

smcmurtry:
To see this bug, delete all the rows from a table like "questionClearLogic" and notice that `yarn download` fails. This PR fixes that.


SupeDeDupe (approved):
🦑👍

PR #2085, opened 2019-04-15
Removed front-end of saved list feature

### LINKED ISSUE #2084: Remove saved list feature - front end
Remove front-end pieces of the saved list feature: the favourites page, save to list buttons on ben cards, saved list buttons on ben dir, saved list icons, etc.

https://app.zeplin.io/project/5acbd659dc553ec84805afd4/screen/5b45ad8f38da59ab5e754fe3

SupeDeDupe:
This PR closes #2084. I only commented out the saved list UI parts and made the favourites page inaccessible on the prod environment. I think it's relevant to keep the backend fragments of the saved list feature for now, as some of it will be reusable in the feature which is to replace it.

There will have to be a thorough cleaning to refine the code: #2086

### DIFF
--- a/__tests__/components/benefit_card_test.js
+++ b/__tests__/components/benefit_card_test.js
@@ -91,7 +91,7 @@ describe("BenefitCard", () => {
 
   it("Clicking the See More button expands the BenefitExpansion component", () => {
     mountedBenefitCard()
-      .find("HeaderButton")
+      .find("BenefitExpansion")
       .at(0)
       .simulate("click");
     expect(mountedBenefitCard().find("BenefitExpansion").length).toEqual(1);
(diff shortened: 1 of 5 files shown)


__tests__/components/benefit_card_test.js
```diff
@@ -91,7 +91,7 @@ describe("BenefitCard", () => {
 
   it("Clicking the See More button expands the BenefitExpansion component", () => {
     mountedBenefitCard()
-      .find("HeaderButton")
+      .find("BenefitExpansion")
```
SupeDeDupe, line 94: This seems to have never been fixed when the See More button was changed to a toggle. The "HeaderButton" should have been replaced with "BenefitExpansion." This whole time, the test was picking up the HeaderButton from the saved list button on the benefit card.

PR #2109, opened 2019-04-23
Removed "Contact Us" button from "no results" message

SupeDeDupe:
Conversation behind removing the contact us button:

Screen Shot 2019-04-23 at 12 47 00 PM

The PR removes the contact us button from the benefits directory page when there are no results. The no_results_button_test was adjusted so it would check for one button, no longer two.


thomascraig (approved):
Approved.

PR #2126, opened 2019-04-30
updated css for IE 11 icon placement

### LINKED ISSUE #2124: Icons need adjusted to render correctly in IE
image

CSS related position issue.

thomascraig:
Closes #2124

### DIFF
--- a/components/vac_header_en.js
+++ b/components/vac_header_en.js
@@ -62,13 +62,13 @@ class VacHeaderEn extends Component {
                     }}
                     className="h6 light-grey"
                   >
-                    Français
+                    <span className="hidden-xs">Français</span>
                     <img
                       id="language-globe"
                       src={
                         path + "/GCWeb/assets/2018-redesign/vector-globe.svg"
                       }
-                      alt=""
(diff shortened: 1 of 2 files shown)


components/vac_header_en.js
```diff
@@ -37,15 +37,16 @@ class VacHeaderEn extends Component {
                   <a
                     href="https://www.canada.ca/en.html"
                     className="h6 light-grey"
+                    rel="noopener"
                   >
-                    Canada.ca
+                    <span className="hidden-xs">Canada.ca</span>
```
SupeDeDupe, line 42: I'm just wondering about the hidden-xs class being added... I'm not sure if hidden-xs exists in our code's css

SupeDeDupe (approved):
Awesome job! I just have a comment about the class hidden-xs

components/vac_header_en.js
thomascraig: span tag and class is used by parent application with our intention to reuse the parent CSS in future builds, this will align the application code with the parent code base.

PR #2128, opened 2019-04-30
Disabled summary page

SupeDeDupe:
The summary page is accessible if the user sets the page to /summary in the url. This patch should disable that for prod.

I noticed the summary page was being hit -- most likely by bots accessing the site -- but it shouldn't be available in prod since the edit selections sidebar was added.

Screenshot 2019-04-30 at 3 41 53 PM


thomascraig (approved):
Good job.

PR #2129, opened 2019-05-01
Hot fixes - disabling summary page + styling fixes to fallback page

### LINKED ISSUE #2132: Fallback page styling
The /all-benefits fallback page should not have the feedback bar on it as its functionality is not available to the browsers that would be redirected to this page. The French link to /all-benefits leads to the English version:

Screenshot 2019-05-01 at 1 31 28 PM

SupeDeDupe:
This PR disables the summary page on production environment, which should never have been accessible to users at this time. Also closes #2132, addressing some small styling things and temporarily removes the header and footer from /all-benefits, so IE10 users don't see the header and footer without styling.

### DIFF
--- a/components/layout.js
+++ b/components/layout.js
@@ -82,13 +82,15 @@ class Layout extends Component {
           <ErrorBoundary>
             <Content>
               <SkipToMainContent skipLink={skipLink} t={t} />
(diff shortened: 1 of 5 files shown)


thomascraig (approved):
approved

PR #2133, opened 2019-05-01
Fix for ie10 header/footer css bug

### LINKED ISSUE #2130: IE10 not reading the header and footer css files
The VAC header and footer styling for IE 10 or lower needs work to match the design.

Screenshot 2019-05-01 at 10 30 00 AM

smcmurtry:
Closes #2130. The fix was to tell the server not to send the "not supported" html page in response to requests for css files.


SupeDeDupe (approved):
Awesome! 🦑 👍

PR #2134, opened 2019-05-03
IE10 header + footer fixed

### LINKED ISSUE #2130: IE10 not reading the header and footer css files

SupeDeDupe:
Closes #2130.

Copied from #2133 because the new version release implementation is not allowing in branches that were created off of the master branch from before this newest setup was applied.


SupeDeDupe:
Removed the CircleCI build job that was holding back PR tests from passing.

PR #2141, opened 2019-05-06
Moved Yes/No in feedback bar

### LINKED ISSUE #1969: Close up the space between "Did you find the information you’re looking for?" And the Yes and No options
Right now the Yes and No are floating a bit too much to the right :) We can close that up, maybe half that space so it looks more connected to the Did you find the information you're looking for? question.

apmcguigan:
closes #1969


SupeDeDupe (approved):
Good job! 🦑👍

PR #2144, opened 2019-05-08
Removed map page references

### LINKED ISSUE #2057: Removed remnants of the map page
Comb through the code and Airtable and the code to ensure no references to /map exist anymore. The getMapUrl function still exists in urls.js, and the areaOffices sheet is unused in Airtable.

SupeDeDupe:
Closes #2057

- Removed variables in codebase which were only used on the map page
- Removed variables on AirTable which were only used on the map page in the translations sheet [this will make it into the code during the next content update]

### DIFF
--- a/selectors/urls.js
+++ b/selectors/urls.js
@@ -121,21 +121,3 @@ export const getPrintUrl = createSelector(
     return "/print?" + params.join("&");
   }
 );
-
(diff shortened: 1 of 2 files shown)


thomascraig (approved):
approved.

PR #2145, opened 2019-05-08
Alpha references removed

### LINKED ISSUE #2102: Refactor AlphaBanner file names/references
The beta banner is in alpha_banner.js file. This should be renamed to something more broad like version_banner, as it is currently the beta banner, but may change to a release banner. All instances of alpha should be removed from the code because we are long past alpha phase.

SupeDeDupe:
Closes #2102.

- Removed old alpha variables
- Renamed alpha_banner.js to version_banner.js
- Renamed AlphaBanner component to VersionBanner


thomascraig (approved):
Approved.

PR #2158, opened 2019-05-16
removed unused dependecies

apmcguigan:
ran `npx depcheck`

Shows all unused dependencies, removed them from package.json just to clean up the project


SupeDeDupe (commented):
To apply these dependency changes to the project, I believe you'll have to run `yarn` to update the yarn.lock file.

PR #2173, opened 2019-06-21
Upgraded jest-haste 

SupeDeDupe:
Upgraded the jest-haste and webpack dependencies to resolve security vulnerability reported by snyk.
Screenshot 2019-06-21 at 2 24 34 PM


SupeDeDupe:
Looks like next@8.1.0 is using webpack@4.32.2 or lower. Next JS is currently at version 8.1.0, so that can't be upgraded but maybe the webpack it uses in several places can be....

PR #2176, opened 2019-08-06
[WIP] upgraded dependency next

SupeDeDupe:


SupeDeDupe:
The security issues aren't being flagged anymore since I upgraded some of the modules, but now the build is broken and it has something to do with importing the SVGs. It may be worth refactoring how we use the SVGs. I'm not sure if the current method is so intuitive...

PR #2179, opened 2019-08-08
Dependency upgrade

SupeDeDupe:
This PR is to patch security vulnerabilities, particularly upgrading lodash, which is used frequently as a sub-dependency.


thomascraig (approved):
Approved.

PR #2181, opened 2019-08-13
Updated intro text design on Guided Experience pg 1

SupeDeDupe:
Shortened margins, changed intro text content, removed grey background.

image


thomascraig (approved):
Good Job

PR #2188, opened 2019-08-16
Mobile Results page styling

SupeDeDupe:
Fixed styling to match design on Results page for mobile:

image


thomascraig (approved):
Nice work. Approved.

PR #2190, opened 2019-08-16
New benefits sorting

SupeDeDupe:
Sorts the benefits using the benefit's `sortOrder` attribute in AirTable. It uses a numbering system, rating 1 as the highest sort priority. Benefits without a `sortOrder` number receive the rating of an arbitrary high number constant -- currently set to 5000, causing it to be ranked at the bottom.


apmcguigan (approved):
Looks great!

GBGRAVES:
Awesome job!

PR #2192, opened 2019-08-20
Search bar analytics

SupeDeDupe:
Added a logEvent to the search bar onChange event, and a state variable to keep track of the analytics having been sent already for the current user session.


thomascraig (approved):
Nice work.

SupeDeDupe:
[Add log event to allow analytics on the search bar](https://trello.com/c/wgtbFvzy/227-add-log-event-to-allow-analytics-on-the-search-bar)

PR #2199, opened 2019-08-21
Remove X btn on modal and rename sidebar variables

SupeDeDupe:
The X button in the Edit Selections Modal was saving responses, which was not desired behaviour. The saving questions function was too strongly coupled to the radio button and checkbox components for an easy fix without requiring major code refactors. The final design decision was to remove the X.


thomascraig (approved):
Nice work.

PR #2200, opened 2019-08-21
Fix edit selections modal header and size

SupeDeDupe:
The edit selections modal's size was too big for the page and scroll didn't work. It was dysfunctional on mobile, and the close button's position was getting pushed below the header.

This PR:

- adjusts the size of the modal on mobile
- allows the scroll
- fixes the close button position to always be in horizontal line with the header
- sticky's the header on scroll, so the close button is always clickable on screen


apmcguigan (approved):
Looks good!

PR #2202, opened 2019-08-22
Search bar clear icon aria label

SupeDeDupe:
Created new aria label for clear button in search bar, because it was originally being mistakenly marked as "search." It's now distinguished as "clear."

If you hover over the icons on the search bar, the X clear icon should now say "clear" and the search icon should say "search."


thomascraig (approved):
Nice work.

PR #2209, opened 2019-09-12
Slight change to Content Update doc

SupeDeDupe:
Removed the md checkboxes that weren't displaying as expected


apmcguigan (approved):
👍 👍

PR #2222, opened 2019-10-29
Removed feedback bar

apmcguigan:


GBGRAVES:
[Disable top feedback mechanism page and remove lead-in text from the tool](https://trello.com/c/xmcZ8NbW/270-disable-top-feedback-mechanism-page-and-remove-lead-in-text-from-the-tool)

apmcguigan:
Removed in #2224

PR #2223, opened 2019-10-29
Benefit Card title now links

apmcguigan:


apmcguigan:
Cleaned up changes and resubmitted as PR #2225

PR #2233, opened 2020-03-03
Removed empty translation

apmcguigan:


thomascraig (approved):
fixes broken dev/lab instance.

